PR #456, opened 2026-06-25
fix(mta-sts): downgrade WAF-challenged policy fetch to inconclusive (#455)

### LINKED ISSUE #455: Scanner WAF-challenged on Cloudflare-fronted origins → false-positive MTA-STS `high` + inconclusive http_security
## Summary

When the scanner audits a **Cloudflare-fronted origin**, Cloudflare's WAF / bot-fight returns an **HTTP 403 challenge page** to the scanner's own fetch. The scanner reads the challenge page instead of the real response, producing two bad outcomes:

1. **`http_security` → `error` / inconclusive** (graceful, acceptable degrade), and
2. **`mta_sts` → false-positive `high` finding** ("MTA-STS policy file not accessible … returned HTTP 403") on a domain whose policy is actually served correctly with `200`.

Symptom #2 is the more serious: it's a **confident wrong answer** that drops the `mta_sts` category to 75 and emits a scary `high`-severity finding on a *correctly configured* domain.

## Reproduction

Target: `blackveilsecurity.com` (M365 mail, Cloudflare-fronted; `enterprise_mail` profile).

`scan_domain` / `check_mta_sts` reports:
```
mta_sts: score 75, severity "high"
"MTA-STS policy file at https://mta-sts.blackveilsecurity.com/.well-known/mta-sts.txt returned HTTP 403."
```
`check_http_security` reports:
```
checkStatus: "error"
"Cloudflare WAF challenge intercepted … httpStatus 403, inconclusive: true"
```

But the policy file **is** served correctly — every non-scanner client gets `200`:
```
$ curl -s -o /dev/null -w "%{http_code}\n" -A "Mozilla/5.0"        https://mta-sts.blackveilsecurity.com/.well-known/mta-sts.txt  → 200
$ curl -s -o /dev/null -w "%{http_code}\n" -A "Go-http-client/2.0" https://mta-sts.blackveilsecurity.com/.well-known/mta-sts.txt  → 200
$ curl -s -o /dev/null -w "%{http_code}\n" -H "User-Agent:"        https://mta-sts.blackveilsecurity.com/.well-known/mta-sts.txt  → 200

$ curl -s -A "Mozilla/5.0" https://mta-sts.blackveilsecurity.com/.well-known/mta-sts.txt
version: STSv1
mode: enforce
mx: blackveilsecurity-com.mail.protection.outlook.com
max_age: 604800
```
Generic automation (Go HTTP client, empty UA) is **not** challenged — only the scanner is. So the WAF rule keys on the scanner's specific request signature (egress IP range and/or UA), not on "is a bot."

## Impact

- Cloudflare fronts a large share of the public web, so this misfires on a substantial fraction of real scans.
- The `mta_sts` false-positive is a **confident high-severity finding on a healthy domain** — it erodes trust in the scanner and would push an operator to "fix" something that isn't broken.
- It also hides genuinely broken MTA-STS behind the same 403, since a real "policy 403" and a "scanner-got-challenged 403" are indistinguishable in the output.

## Suggested direction

For any **HTTPS-fetch-based check** (`http_security`, `mta_sts` policy fetch, BIMI/SVG, security.txt, etc.):

1. **Detect the CF challenge** (the code already recognizes `wafKind: "challenge"` for `http_security`) and apply the **same inconclusive/degrade path to `mta_sts`** — never emit a confident `high` "not accessible" when the 403 is a WAF challenge rather than a real origin error. Downgrade to `inconclusive` + a `representative`/`missingControl: false` note instead.
2. **Reduce the challenge rate**: set a stable, identifiable scanner User-Agent and document the scanner egress so CF-fronted origins (incl. ours) can allowlist it; consider a retry with that UA before concluding "inaccessible."
3. Distinguish **"policy fetch challenged" vs "policy genuinely 4xx/5xx"** in the finding text so operators aren't told a working policy is broken.

The fetch + WAF-detection logic likely lives in `packages/dns-checks` (or the worker's fetch wrapper) — maintainers to route.

## Environment

- Scanner: hosted bv-mcp (`dns-mcp.blackveilsecurity.com`), scoringModelVersion 1.2.0, configHash 8393a61c.
- Date observed: 2026-06-25.
- Reproduction domain `blackveilsecurity.com` is itself Cloudflare-fronted (served by bv-web-prod Worker), policy served via the `.well-known/mta-sts.txt` route.

MadaBurns:
Fixes #455.

## Problem

A Cloudflare/Akamai WAF challenge or block on the **MTA-STS policy fetch** (commonly HTTP 403) made `check_mta_sts` emit a confident `high` **"MTA-STS policy file not accessible"** finding — a false positive on a policy that real MTAs can fetch fine. Reproduction: `blackveilsecurity.com` serves `mode: enforce` at `200` to every non-scanner client (`Mozilla`, `Go-http-client`, empty-UA), but the scanner's own probe is WAF-challenged → `403` → a scary `high` on a healthy domain, dropping the category to 75.

Same root cause as the inconclusive `http_security` result on CF-fronted origins — `check_http_security` already detects the challenge and degrades gracefully; `check_mta_sts` did not.

## Fix

Mirror the WAF detection already used by `check_http_security`, at the **same (worker) layer** — not the package — so the portable `@blackveil/dns-checks` core stays pure and there's no parity-contract / publish / re-vendor churn:

- **`src/lib/waf-detection.ts`** (new) — extract the shared `looksLikeWaf` / `detectWafEvent` helpers (were private to `check-http-security.ts`). No behavior change.
- **`src/tools/check-http-security.ts`** — import the extracted helpers.
- **`src/tools/check-mta-sts.ts`** — wrap the policy `fetch` to **observe** the response; on a detected challenge/block, replace the `high` finding with an inconclusive `info` and recompute the score via `buildCheckResult`. Observation is **fail-open**: it never alters the response the package sees, so a genuine non-WAF 4xx still surfaces the `high`.

## Tests

`test/check-mta-sts-waf.spec.ts` (new):
- CF challenge (`403` + `cf-mitigated: challenge`) → no `high`, inconclusive `info`, control still credited, score recovered.
- CF block (`403` + block body) → inconclusive `info`, `wafKind: block`.
- Plain `nginx` `403` (no CF/Akamai signals) → `high` **preserved** (guards against over-broad suppression).

All green locally: new spec 3/3; existing `check-mta-sts.spec.ts` 15/15 (no regression); `check-http-security`, `mta-sts-analysis`, `handlers-tools`, `format-scan-report` pass. Typecheck + ESLint clean on changed files.

## Tradeoff to review

Downgrading a challenged 403 to *inconclusive* means a domain that genuinely blocks **all** automation (incl. real MTAs) now reads "inconclusive" rather than "high failure." Mitigations: the finding is explicitly inconclusive (never a pass), and TLS-RPT reports remain the authoritative real-delivery signal. The score does recover off the 75 floor (info findings are non-penalizing in the model) — whether *inconclusive* should score that high is a judgment call; left as-is and flagged here rather than deepening the scoring model in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ## [Unreleased]
 
+## [3.25.4] - 2026-06-25
+
+Patch release: one **server-only** `check_mta_sts` false-positive fix. No `@blackveil/dns-checks` core change, `SCORING_MODEL_VERSION` unchanged, tool count unchanged (79). Normal-corpus scores are unaffected — the new behavior only applies when the MTA-STS policy fetch is intercepted by a WAF.
+
+### Fixed
+
+- **`check_mta_sts` no longer emits a confident `high` "policy file not accessible" on a WAF-challenged policy fetch (#455).** The `mta-sts.<domain>` policy host is often Cloudflare/Akamai-fronted, and the scanner's own probe gets a WAF challenge/block (commonly HTTP 403) while real sending MTAs fetch the policy fine — so a healthy `mode: enforce` policy (e.g. `blackveilsecurity.com`) was falsely flagged. The `check_mta_sts` worker wrapper now detects the interception (reusing the `check_http_security` WAF fingerprint, extracted to `src/lib/waf-detection.ts`) and downgrades the finding to an inconclusive `info` with a fixed `penaltyOverride` (lands ~90 — off the ceiling since the policy is unverifiable, but above a real failure since the `_mta-sts` TXT record was verified). The body sniff is bounded (8 KB) so an attacker-controlled policy host can't force a large-body read. A genuine non-WAF 4xx still surfaces the `high`.
+
 ## [3.25.3] - 2026-06-24
 
 Patch release: ships the **finding-sanitizer hardening** merged in #438 to prod. Bundles **`@blackveil/dns-checks` 1.3.18**. `SCORING_MODEL_VERSION` unchanged, scores unchanged (string-channel sanitization only), tool count unchanged (79).
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "blackveil-dns",
-	"version": "3.25.3",
+	"version": "3.25.4",
 	"lockfileVersion": 3,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "blackveil-dns",
-			"version": "3.25.3",
+			"version": "3.25.4",
 			"license": "BUSL-1.1",
 			"workspaces": [
 				"packages/*"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "blackveil-dns",
-	"version": "3.25.3",
+	"version": "3.25.4",
 	"license": "BUSL-1.1",
 	"type": "module",
 	"bin": {
--- a/server.json
+++ b/server.json
@@ -6,7 +6,7 @@
 		"url": "https://github.com/MadaBurns/bv-mcp",
 		"source": "github"
 	},
-	"version": "3.25.3",
+	"version": "3.25.4",
 	"remotes": [
 		{
 			"type": "streamable-http",
--- a/src/lib/response-body.ts
+++ b/src/lib/response-body.ts
@@ -8,3 +8,95 @@ export async function disposeUnreadResponseBody(response: Response): Promise<voi
 		// The body may already be locked or consumed.
 	}
 }
+
+/**
+ * Bounded streaming body readers.
+ *
+ * Both read at most `maxBytes` BYTES from a response body stream, then cancel
+ * the reader (always, in a `finally`). The bound is on the cumulative BYTE
+ * length of the chunks read (`Uint8Array.byteLength`) — NOT the decoded string
+ * length, which on multi-byte UTF-8 can be ~3-4x smaller than the byte count
+ * and would let an attacker-controlled endpoint buffer far more than intended.
+ *
+ * Fail-open: never throw. The body stream the caller passes is the only thing
+ * consumed — the caller owns `response.clone()` if it needs the original
+ * undisturbed.
+ */
+
+/** Drain the stream up to a byte cap. Returns the accumulated chunks, the
+ * cumulative byte total read, and whether the cap was reached/exceeded. */
+async function drainBounded(
+	body: ReadableStream<Uint8Array> | null,
+	maxBytes: number,
+): Promise<{ chunks: Uint8Array[]; total: number; overflowed: boolean }> {
+	const chunks: Uint8Array[] = [];
+	let total = 0;
+	let overflowed = false;
+	if (!body) return { chunks, total, overflowed };
+	const reader = body.getReader();
+	try {
+		for (;;) {
+			const { done, value } = await reader.read();
+			if (done) break;
+			if (!value) continue;
+			chunks.push(value);
+			total += value.byteLength;
+			if (total >= maxBytes) {
+				overflowed = true;
+				break;
+			}
+		}
+	} finally {
+		// Always release the underlying stream — best-effort, swallow errors.
+		try {
+			await reader.cancel();
+		} catch {
+			/* fail-open */
+		}
+	}
+	return { chunks, total, overflowed };
+}
+
+/** Concatenate chunks into one Uint8Array. */
+function concatChunks(chunks: Uint8Array[], total: number): Uint8Array {
+	const out = new Uint8Array(total);
+	let offset = 0;
+	for (const c of chunks) {
+		out.set(c, offset);
+		offset += c.byteLength;
+	}
+	return out;
+}
+
+/**
+ * Read at most `maxBytes` bytes from a body stream, then cancel it. Byte-accurate
+ * (bounds on cumulative `Uint8Array.byteLength`, truncating the body once the cap
+ * is reached). Fail-open: returns '' on any error or a null/empty body. Caller
+ * owns cloning the response if it needs the original undisturbed.
+ */
+export async function readBoundedText(body: ReadableStream<Uint8Array> | null, maxBytes: number): Promise<string> {
+	try {
+		const { chunks, total } = await drainBounded(body, maxBytes);
+		if (total === 0) return '';
+		return new TextDecoder().decode(concatChunks(chunks, total));
+	} catch {
+		return '';
+	}
+}
+
+/**
+ * Like {@link readBoundedText} but NULL-on-overflow: returns `null` when the body
+ * exceeds `maxBytes` (rather than truncating), and `null` on any error or a null
+ * body. Used where the caller must treat an over-cap body as "unverifiable"
+ * rather than processing a partial prefix (e.g. capability-document integrity).
+ */
+export async function readBoundedOrNull(body: ReadableStream<Uint8Array> | null, maxBytes: number): Promise<string | null> {
+	try {
+		if (!body) return null;
+		const { chunks, total, overflowed } = await drainBounded(body, maxBytes);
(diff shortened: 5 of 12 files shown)


MadaBurns:
## Self-review (independent reviewer pass)

Reviewed the diff with fresh eyes + an independent reviewer. One real issue found and fixed in `757759a`; the rest verified correct.

### 🔴 HIGH (fixed in 757759a) — unbounded attacker-controlled body read
`mta-sts.<domain>` is controlled by the **scanned domain owner**. Because Cloudflare's Worker egress injects `cf-ray` on *every* response (documented in `check-http-security.ts`), `looksLikeWaf` is `true` for every non-ok policy fetch in production — so `sniffBody` ran on every 4xx/5xx. The original `response.clone().text()` had **no size cap**, so a hostile host could serve a multi-MB 403 body and force the scanner to buffer it (CPU/memory DoS).

**Fix:** bounded streaming read (`MAX_WAF_SNIFF_BYTES = 8 KB`) that cancels the stream after the cap. WAF markers ("just a moment", "you have been blocked") appear in the first bytes, so detection is unaffected. Added a test that serves a hostile ~1 MB 403 body and asserts detection still works without buffering it all, plus a WAF-intercepted 301-redirect-branch test.

### ✅ Verified correct
- **Score recompute / `controlPresent`** — `controlPresent` derives from `hasTxtRecord` (the `_mta-sts` record was observed), independent of the policy-file fetch, so preserving it is right. The inconclusive finding carries `inconclusive: true` but **not** `missingControl: true`, so the score computes normally off the kept findings rather than being forced to 0.
- **Suppression scope** — the title set covers exactly the two findings a WAF 403/3xx can falsely trigger (`policy file not accessible`, `policy redirects`); `oversized` / `fetch failed` can't be triggered by a WAF 403 and are correctly left out. A genuine non-WAF 403 (no CF/Akamai signal) is **not** downgraded (tested).
- **Observation invisibility** — fail-open: any error leaves `policyWafEvent` null and the original response is always returned untouched, so the package's behavior is unchanged on the non-WAF path.
- **Refactor** — `looksLikeWaf`/`detectWafEvent`/`WafEvent` extraction matches the `check-http-security.ts` call sites exactly; no behavior change there.

### ⚖️ Open tradeoff for a maintainer (not a blocker)
Downgrading to *inconclusive* lets the score recover off the 75 floor (info findings are non-penalizing). A domain that genuinely blocks **all** automation incl. real MTAs would now read "inconclusive" rather than "high failure." The finding is explicitly inconclusive (never a pass) and TLS-RPT remains the authoritative real-delivery signal — but if you'd prefer *inconclusive* to score lower (neutral, not near-100), that's a scoring-model call worth making here.

Tests: 66 green (`check-mta-sts-waf` 5, `check-mta-sts` 15, `check-http-security` full). Typecheck + ESLint clean on changed files.

MadaBurns:
Resolved the scoring tradeoff in `ebb32c8`: the inconclusive finding now carries `penaltyOverride: 10` (the `check-dnssec` pattern), so a WAF-challenged policy lands **~90** instead of a perfect 100 — off the ceiling (the policy was unverifiable) but above a real `high` failure (75, since the `_mta-sts` TXT record *was* verified). Test locks the band (>75, <100). All green.

MadaBurns:
Parking this in favor of the **concurrent, more complete** #455 fix in progress (uncommitted in the local checkout at the time of writing). That work supersedes this PR — it fixes a real bug here and refactors the shared helpers. Closing to avoid a double-merge.

**What that work already does better (carry it forward):**
- **Akamai false-positive fix** — `detectWafEvent` here treats *any* `Server: AkamaiGHost` response as a WAF block, but AkamaiGHost is on every Akamai response. The concurrent version correctly gates the Akamai branch on `status >= 400` + an access-denied body signature. (This bug was copied verbatim from `check-http-security.ts`, so it pre-dates #455 — worth fixing there too.)
- **Shared `src/lib/response-body.ts`** (bounded read) replacing the inline `sniffBody`, and a shared `buildWafFinding()` so `check_mta_sts` / `check_http_security` don't duplicate the inconclusive-finding construction.
- **`mta-sts-title-coupling.audit.test.ts`** locking `POLICY_FETCH_FALSE_POSITIVE_TITLES` against the package's real emitted finding titles — closes the drift risk this PR left open.

**Two reconciliation notes for the consolidated PR:**
1. **Scoring intent (don't regress):** the inconclusive MTA-STS result should land **~90**, via `penaltyOverride: 10` on an `info` finding — NOT `missingControl: true`. In `buildCheckResult`, `missingControl: true` forces the category score to **0**, which is worse than the original false-positive (75). The concurrent `buildWafFinding` currently sets `missingControl: true` (correct for `http_security`, wrong for the `mta_sts` path) — split that per-caller.
2. **Remote branch state:** `fix/mta-sts-waf-challenge-455` on origin is at `3c89c71` (1 commit ahead of the local `ebb32c8`) — it carries the **release bump to 3.25.4** (package.json/lock, server.json, CHANGELOG). Reuse or cherry-pick that bump for the consolidated release rather than redoing it.

Full review trail (DoS bounded-read, scoring rationale) is in the comments above. Reopen if the consolidation stalls.

MadaBurns:
## Follow-up: addressed the high-effort code-review findings (commit `1fd4c9b`)

A multi-agent review surfaced 10 findings; all are now fixed on this branch. Mapping:

### 🔴 Correctness
- **Akamai branch suppressed genuine failures unconditionally** — `detectWafEvent`'s Akamai path returned `block` for *any* non-ok response with `Server: AkamaiGHost` (a generic edge header), so a real 403/404/500 behind Akamai was mis-attributed to a WAF and silently downgraded. Now gated to `status>=400` + an Akamai access-denied body signature, mirroring the CF branch's rigor. (`src/lib/waf-detection.ts`)
- **Self-inflation / unverifiable check scored ~90** — the bespoke `info + penaltyOverride:10` let an owner-controlled `mta-sts.<domain>` response push the score to ~90. Now **aligned with `check_http_security`**: `checkStatus:'error'` → the category is **excluded** as inconclusive (neither pass, fail, nor inflate). Dropped `INCONCLUSIVE_POLICY_PENALTY`. (`src/tools/check-mta-sts.ts`)
- **Timeout/throw path left uncovered** — a stalled WAF challenge that aborts produces the package's transient `medium` "policy fetch failed", which the title-Set never downgraded. Now the policy-fetch rejection (Abort/Timeout/network) is observed + re-thrown, and the result is excluded the same way. A genuine deterministic 404 still keeps its `high`. (`src/tools/check-mta-sts.ts`)

### 🟡 Robustness
- **Brittle cross-package title coupling** — added an audit tripwire (`test/audits/mta-sts-title-coupling.audit.test.ts`) that fails if `@blackveil/dns-checks` rewords either policy-fetch title, so the downgrade Set can't silently break.

### 🟢 Quality
- Body sniff now gated to `status>=400 || cf-mitigated` (skips clone+read on genuine redirects/404s where `cf-ray` rides every CF egress).
- Extracted shared `buildWafFinding` (carries the WAF metadata contract once; both check tools consume it) and a byte-accurate `readBoundedText` (`src/lib/response-body.ts`) — fixes the prior `MAX_WAF_SNIFF_BYTES` UTF-16-vs-bytes bound and the duplicated finding construction.
- Block events no longer mislabeled "challenge" in the finding title (kind-aware).
- Single-fetch contract guard on the observe wrapper.

**One intentional non-change:** `check-agent-discovery.ts` keeps its own bounded reader — it feeds raw bytes to SHA-256 for capability-doc integrity, and routing through the string util would corrupt non-UTF-8 bytes.

**Note for release:** this changes the WAF scoring treatment described in the 3.25.4 changelog entry (now *excluded/inconclusive* rather than info ~90) — may warrant a changelog/version touch-up.

Verification: `tsc` + `eslint` clean; 247 tests pass across the affected specs (waf-detection, response-body, title-coupling audit, check-http-security, check-mta-sts, check-mta-sts-waf, scan-domain, check-agent-discovery, mta-sts-analysis, handlers-tools, format-scan-report).

PR #508, opened 2026-07-10
ci: add symlink-escape-gate reusable workflow (GhostApproval / CWE-61)

MadaBurns:
## Summary

Adds `.github/workflows/symlink-escape-gate.yml`, a reusable GitHub Actions workflow that fails any push/PR whose tracked files include a symlink resolving outside the repository root (CWE-61).

**Background:** Wiz disclosed ["GhostApproval"](https://www.wiz.io/blog) (2026-07-08) — a malicious repo can ship a symlink such as `config.json -> ../../../.ssh/authorized_keys` that an AI coding agent follows, writing outside the workspace while its confirmation dialog shows the fake in-repo path. bv-mcp is a public repo (npm package `blackveil-dns` + hosted MCP server at `dns-mcp.blackveilsecurity.com`) that already receives external contributions (dependabot today; plausibly real external PRs over time given it's a published product), so this closes that primitive at the CI boundary.

**Why bv-mcp and not bv-cc:** A canonical, locally-runnable detector (`tools/lint-symlink-escape.mjs`) already exists in the private `bv-cc` governance repo for manual/local use. But GitHub does not allow a public repo to consume a `workflow_call` reusable workflow hosted in a private repo — so the actual CI gate has to live somewhere public. bv-mcp already plays this role for two other repos (`bv-claude-dns`, `blackveil-dns-action` both consume `bv-mcp`'s existing `repo-hygiene.yml` via `uses: MadaBurns/bv-mcp/.github/workflows/repo-hygiene.yml@main`), so this workflow follows that exact established pattern — same checkout pin, same trigger shape (`workflow_call` + `push`/`pull_request` on `main`), same self-contained-bash-no-external-fetch design.

**Design notes:**
- Self-contained bash, no external script fetch — so public consumers don't need read access to private `bv-cc`.
- The `bv-cc` copy (`tools/lint-symlink-escape.mjs`) and this workflow are kept in lockstep by hand, not by import — flagging that as a maintenance note, not a blocker.
- Absolute symlink targets (`/opt/homebrew/...`, `/usr/bin/...`) are intentionally NOT flagged — near-universal in real repos (venv interpreters, vendored binaries) and don't generalize across machines/usernames the way relative `../` traversal does, so they aren't the GhostApproval primitive.
- Home-relative (`~/...`) targets are flagged as a symlink-escape.

## Test plan

- [x] YAML validated well-formed via Ruby's built-in `psych` parser (no `pyyaml`/`js-yaml` available locally) — parses cleanly, same `on:` → `true`-key YAML-1.1 quirk as every other workflow in this repo.
- [x] Bash detection logic proven correct against **real GNU coreutils on `ubuntu:24.04` in Docker** (matching the `ubuntu-latest` runner) — not just macOS BSD tools, which lack `realpath -m` entirely:
  - Clean relative symlink (resolves inside a fake repo root) → 0 violations, exit 0.
  - Malicious relative symlink (`../../outside/secret`, escapes a fake repo root) → 1 violation reported via `::error`, exit 1.
- [x] Ran the same scan against bv-mcp's actual working tree. **Zero symlinks are git-tracked** in this repo today (`git ls-files -s | awk '$1==\"120000\"'` → empty) — the filesystem does have a few symlinks (a client-internal pair under `.dev/`, `.venv/bin/python*`), but all are covered by `.gitignore` and would never reach a fresh `actions/checkout`, so this gate is a no-op on current `main` and only activates against future tracked symlinks.
- [x] **Known limitation found during verification, documented here rather than silently patched (task scope was logic-frozen):** `realpath -m` fully dereferences *existing* intermediate symlinks it encounters while resolving a path — it only skips resolution for a *final missing* component. So a benign same-directory relative symlink whose target is itself an existing symlink to an absolute external path (e.g. the classic venv chain `python3 -> python -> /opt/homebrew/.../python3.13`) gets misresolved through that chain and can spuriously read as escaping via `../`. Reproduced directly against `.venv/bin/python3` in this repo. Doesn't affect this PR (the whole `.venv/` is gitignored, so it's invisible to CI), but worth a follow-up if the gate ever needs to cover a tracked chained-symlink layout.

### DIFF
--- a/.github/workflows/symlink-escape-gate.yml
+++ b/.github/workflows/symlink-escape-gate.yml
@@ -56,12 +56,7 @@ jobs:
             # `realpath` would wrongly dereference any *existing* intermediate
             # symlink and conflate a harmless chain (e.g. venv's
             # bin/python3 -> python -> /abs/real/python) into a false escape.
-            rel=$(python3 -c "
-import os, sys
-root, link, target = sys.argv[1], sys.argv[2], sys.argv[3]
-resolved = os.path.normpath(os.path.join(os.path.dirname(link), target))
-print(os.path.relpath(resolved, root))
-" "$ROOT" "$link" "$target")
+            rel=$(python3 -c 'import os, sys; root, link, target = sys.argv[1], sys.argv[2], sys.argv[3]; resolved = os.path.normpath(os.path.join(os.path.dirname(link), target)); print(os.path.relpath(resolved, root))' "$ROOT" "$link" "$target")
             case "$rel" in
               ..|../*)
                 echo "::error file=$link::SYMLINK_ESCAPE — relative target escapes repository root via ../ traversal (-> $target)"
--- a/.github/workflows/symlink-escape-gate.yml
+++ b/.github/workflows/symlink-escape-gate.yml
@@ -1,0 +1,67 @@
+name: Symlink escape gate
+
+# Fail PRs/pushes whose tracked files include a symlink resolving outside the
+# repository root (CWE-61 — the primitive behind "GhostApproval", Wiz
+# 2026-07-08: a malicious repo ships e.g. config.json -> ../../../.ssh/authorized_keys,
+# and an AI coding agent's confirmation dialog shows the fake in-repo path
+# while the write lands outside the workspace). Self-applied (push/pull_request)
+# and reusable (workflow_call) — self-contained bash, no external script fetch,
+# so public repos can consume it without needing read access to private bv-cc
+# (the canonical, locally-runnable copy of this check lives there as
+# tools/lint-symlink-escape.mjs — kept in lockstep by hand, not by import).
+on:
+  workflow_call:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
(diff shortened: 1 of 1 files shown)


MadaBurns:
Follow-up fix pushed: the original \`realpath -m\` fully dereferences any *existing* intermediate symlink along the path (it only tolerates a missing *final* component), which broke the intended single-hop judgment — a harmless same-directory chain like \`venv/bin/python3 -> python -> /opt/homebrew/.../python3.13\` got misresolved to its final absolute target and misreported as escaping the repo before the \`/*\` early-exit check ever saw it.

Replaced with a \`python3 -c\` one-liner (\`os.path.normpath\`/\`os.path.relpath\` — pure string manipulation, no filesystem access), so each symlink is judged only by its own immediate target text — exactly matching \`tools/lint-symlink-escape.mjs\`'s \`path.resolve\`/\`path.relative\` in bv-cc. Verified against a synthetic tree covering the multi-hop chain, a real \`../\` traversal escape, and a \`~\`-prefixed target: only the two genuine attacks are flagged, the chain is silent. This also drops the GNU-coreutils \`realpath -m\` dependency (portable python3, no more BSD/GNU divergence to worry about).

PR #551, opened 2026-07-25
fix: two silently-empty internal-call paths (registrar-complement deep-scan envelope, tenant scan capture)

MadaBurns:
Two independent defects with the same failure mode: a consumer reads a field or hook the producer never populates, so the feature reports nothing instead of erroring — and the tests passed because they fabricated the missing shape themselves.

## 1. Registrar-complement deep-scan read `structured`, a field the internal call never emits (`924e13c`)

`runDeepScan` unwrapped each internal tool call as `envelope.structured`, but the production `internalCall` closure (`src/index.ts`) returns `handleToolsCall`'s MCP result — `{ content, structuredContent?, isError? }`. There is no `structured` field on that path, so `r.scan` was `undefined` for every apex, the `!r.scan` guard skipped all of them, and the job wrote a `registrar_complement_full` payload with `apexesScanned: 0` and an empty grade distribution instead of failing. Operator-deploy only (requires `BRAND_AUDIT_QUEUE`).

Two further shape mismatches on the same path, which would have left `danglingDns` permanently empty even after the rename:

- `scan_domain`'s `structuredContent` is `StructuredScanResult`: `categoryScores` holds plain numbers (or `null`), not `{score, findings}`, and the payload carries no per-finding array at all — only `findingCounts`. Dangling-DNS detail now comes from a per-apex `check_subdomain_takeover` call, whose `structuredContent` IS a `CheckResult` with real findings.
- `Finding` has no `subdomain` field, so the dangling FQDN and CNAME target are recovered from the finding's title/detail text (`metadata` preferred if a future package version emits it).

Also: a failed `discover_subdomains` / `check_subdomain_takeover` is now a per-section partial that preserves the apex's posture rather than dropping the apex (matching the module docstring), and an `isError` envelope is treated as a failed call rather than a zero-value result.

## 2. `scan_domain` never fired `resultCapture`, so every tenant scan row persisted nulls (`aa5d4e7`)

Both tenant scan paths (`src/tenants/queue-consumer.ts`, `src/tenants/routes.ts`) collected their `scan_domain` result via `ToolRuntimeOptions.resultCapture`. That hook fires only on the single-`CheckResult` `TOOL_REGISTRY` path (and `check_resolver_consistency`); `scan_domain` is an orchestrator returning a `ScanDomainResult` and never invoked it, so `captured` was null on every scan. Four silent consequences:

- every `scans` row landed with null `score`, null `grade`, null `result_json`, `finding_count` 0
- nothing ever reached the `findings` table
- the tenant report endpoint (`SELECT score, grade FROM scans WHERE cycle_id = ?`) had nothing to aggregate
- the Phase-6 fingerprint pre-flight gates on `if (lastScan.result_json)` — always false, so the skip-unchanged-domain optimisation could never engage

Adds a `scanResultCapture` hook invoked by the `scan_domain` case, plus `src/tenants/scan-snapshot.ts` as the single definition of the compact projection (`score` / `grade` / `maturityStage` / `findings`) that `result_json` is written from and parsed back into, so the two directions cannot drift. `maturity_stage`, previously hardcoded `null`, is now populated. `grade` is the canonical 9-band `ScanScore.grade`; display surfaces derive the NIST letter from the stored numeric score. **No migration needed** — the column has only ever held nulls.

## Tests

The specs previously injected `{ content: [], structured: {…} }` and called `resultCapture` themselves with a fabricated `CheckResult` — shapes production never emits — so they passed while production produced nothing.

- the registrar-complement deep-scan spec now builds every envelope with the real production emitters (`buildToolResult` wrapping `buildStructuredScanResult`, a real `SubdomainDiscoveryResult`, and a genuinely package-produced `CheckResult` from `checkSubdomainTakeover` with an injected resolver), plus an explicit negative test pinning that a `structured`-keyed envelope yields nothing.
- `test/scan-result-capture-wiring.spec.ts` (new) calls the **real** `handleToolsCall` with DNS mocked and pins which hook fires for `scan_domain` vs a registry check, plus the `result_json` round-trip.
- The tenant mocks emit through a shared `test/helpers/scan-capture.ts` using the real hook and a `ScanDomainResult`-shaped payload.
- Both fixes were verified to fail without their change: suppressing the emitter fails 2 wiring tests; reverting the consumer fails 4 tenant tests.

## Notes for reviewers

- **Fan-out**: the deep-scan now issues a third call per apex. Cold-cache worst case is ~25 apexes × (scan_domain ~20 + discover ~2 + takeover ~15–30) subrequests in one queue invocation — comfortably under the Paid-plan 10,000 ceiling, but it is a real increase.
- **Pre-existing, not addressed here**: a DLQ-written `result_json` still satisfies the `routes.ts` fingerprint pre-flight guard and can skip a scan for up to 24h. Control flow there is identical before and after this PR; the queue-consumer path is strictly improved (an unusable payload now falls through to a real scan instead of reusing garbage).
- **CI**: `contract` is red at `npm audit --audit-level=high` on dev-dependency advisories (`sharp`/`miniflare`/`wrangler`/`brace-expansion`). It fails identically on `main` at this PR's base `0b8b2911` and has since 2026-07-23 — see the [comment thread](https://github.com/MadaBurns/bv-mcp/pull/551#issuecomment-5081255354). Not required, not introduced here; the fix is a toolchain bump that belongs in its own PR.

https://claude.ai/code/session_01UxVFp4AMEQCpCCk4SPWQEJ


MadaBurns:
## CI status: 14/15 green — the one red check is failing on `main` too

`contract` (`ci-contract.yml`) is red, but **not because of this diff**. The contract/Zod test step inside that job passes; the job exits 1 at its last step:

```
Run npm audit --audit-level=high
sharp <0.35.0  — GHSA-f88m-g3jw-g9cj (CVE-2026-33327/33328/35590/35591)
  miniflare → @cloudflare/vitest-pool-workers, wrangler
brace-expansion <=5.0.7 — GHSA-mh99-v99m-4gvg
5 high severity vulnerabilities
```

All dev dependencies, none introduced here.

**It reproduces on the base branch.** `ci-contract.yml` has failed on every `main` push since 2026-07-23 — including [`0b8b2911`](https://github.com/MadaBurns/bv-mcp/commit/0b8b29114ac24aa281eaf108ebd3a0107dff7601), this PR's exact base ([run 30072868810](https://github.com/MadaBurns/bv-mcp/actions/runs/30072868810)) — at the identical `npm audit` step with the identical `sharp`/`miniflare`/`wrangler` chain. The only delta here is `brace-expansion`, an advisory that landed in the tree since (4 vulns on main → 5 now). Last green `contract` run on `main` was 2026-07-21.

Worth noting the two audit gates disagree by design, which is why `Dependency audit` is green while this is red:

| Workflow | Command | Result |
| --- | --- | --- |
| `security.yml` → `Dependency audit` (required) | `npm audit --omit=dev --audit-level=moderate` | ✅ |
| `ci-contract.yml` → `contract` (not required) | `npm audit --audit-level=high` | ❌ |

`contract` is not in the required set (`build-and-test`, `Secret & PII scan`, `Dependency audit`, `File hygiene check`), which matches `mergeStateStatus: unstable` rather than `blocked`.

**Not fixing it here.** The remedy is `npm audit fix`, which bumps `wrangler` / `miniflare` / `@cloudflare/vitest-pool-workers` — the runtime every test executes in. That's a toolchain upgrade that deserves its own PR and its own full-suite verification against `main`, not a silent stack on top of an unrelated bug fix.

### One thing for the author

The branch now carries a **second commit** the PR title and body don't describe — `aa5d4e7`, a separate silent-empty defect in the tenant scan paths (`scan_domain` never fires `resultCapture`, so every `scans` row persisted null `score`/`grade`/`result_json`, no `findings` rows, and the Phase-6 fingerprint pre-flight could never engage). Same failure class as the deep-scan bug, different subsystem. Happy to update the PR description to cover both, or split it out — say which you'd prefer.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #562, opened 2026-07-26
fix(discover-subdomains): make CT enumeration resilient to source outage (failover + stale fallback + per-source health)

MadaBurns:
## Summary

`discover_subdomains` was effectively non-functional whenever crt.sh was unreachable — returning `Certificate Transparency source unavailable` with zero subdomains for hours at a time, for domains that demonstrably have many live subdomains (observed against `bnz.co.nz`, 2026-07-26).

**Root cause, confirmed rather than assumed:** the direct (non-`BV_CERTSTREAM`) enumeration path in `src/tools/discover-subdomains.ts` depended on a *single* public CT source. The Certspotter endpoint referenced in the issue lives in the separate bv-certstream-worker, so it was never reachable from this path — any crt.sh timeout, 5xx, or network error went straight to `emptyResult(domain, true)`. There was no failover in practice.

This is the follow-on half of the earlier bv-web-prod fix that made the tool fail *loudly*. That fix works (the tool now errors instead of returning a passing empty), but the source itself was down, so enumeration returned nothing regardless.

### Three layers of resilience

1. **Real multi-source failover** — crt.sh → Certspotter, each normalized into the same aggregation and tried in order until one answers authoritatively, with a bounded same-source retry plus backoff on a transient outcome. This is deliberately *sequential first-responder* rather than fan-out-and-merge: querying every source cannot fit the 24s synchronous budget (`DISCOVER_SUBDOMAINS_SYNC_BUDGET_MS`). The trade-off is documented at `queryDirectSources`.
2. **Last-known-good fallback** — a clean enumeration is cached per domain (7-day TTL, `SCAN_CACHE` KV) and re-served on a total outage marked `stale: true` with `cacheAgeMinutes` and an explicit staleness banner. A flagged stale list is far more useful to a security caller than an empty result. Also consulted when the budget trips before a direct source can run, since a KV read costs milliseconds unlike the 10s fetch there's no longer budget for.
3. **Per-source observability** — every attempt emits a `ct_source` log line carrying the source and its outcome (`ok`/`empty`/`http_error`/`timeout`/`error`), so upstream outages are visible and alertable instead of silently degrading the tool.

The explicit `sourceUnavailable` error (`isError: true`) is now reached **only** when every live source failed *and* no cache exists. A genuine "queried, none found" answer stays distinct from both. `force_refresh` still bypasses the certstream fast-path cache but no longer suppresses the stale fallback — a total outage must never return a passing empty result.

### Two defects found while reviewing my own implementation

Both are fixed and covered by tests:

- A stale **empty** result was rendered as a confident `no subdomains found in Certificate Transparency logs`, because the zero-subdomain branch in the formatter returned *before* the staleness check — reproducing the exact "unreliable answer read as authoritative" confusion this PR exists to fix. Empty enumerations are now never stored as last-known-good (an empty stale set tells the caller nothing the outage error doesn't), and the zero branch marks staleness defensively.
- A deadline trip before any direct source ran returned the empty error without consulting the cache.

## Type of change

- [x] Bug fix

## Security impact

Touches outbound fetch and caching, so worth a look:

- **New outbound host `api.certspotter.com`** — a hardcoded literal hostname, not attacker-controlled, using `redirect: 'manual'`, matching the existing crt.sh pattern. Per CLAUDE.md, `safeFetch` is required for *attacker-controlled* URLs (BIMI `l=`, redirect `Location` targets); a fixed host with an already-`validateDomain()`'d domain interpolated via `encodeURIComponent` into a query param may use raw `fetch`. No new SSRF surface.
- **DoS bounds preserved on the new source** — the Certspotter fetch reuses the same `content-length` pre-check plus 5 MiB streaming ceiling (`readBoundedOrNull`) as crt.sh, and its entries flow through the same `MAX_SUBDOMAINS` (100) and 5000-entry caps.
- **No new trust surface from CT data** — Certspotter names go through the identical `buildResultFromEntries` filtering that requires each name to end with the target domain suffix, so threat-model item T75.T (malicious CT data inflating candidate sets) keeps its existing mitigation. This PR strengthens T76.D (*"CT source outage degrading discovery"*), which the threat model previously listed as mitigated by a crt.sh fallback that did not actually exist on this path.
- **Cache-poisoning guard** — the LKG entry is written only from a clean success; never from a `sourceUnavailable`, `partial`, already-`stale`, or empty result. Key is `cache:subdomains-lkg:<validated-domain>`; TTL (7 days) is well above the 60s KV floor.
- **No new client-visible error strings**, so the `sanitizeErrorMessage` safe-prefix allowlist is unaffected. Subdomain names in output still route through `sanitizeOutputText`; the staleness banner is our own prose plus a computed integer age.
- **No log-injection vector** — `logCtSource` passes `domain` as a top-level field, which `logEvent` explicitly runs through `sanitizeString`; `source` sits in `details` (routed through `sanitizeLogValue`) and is a hardcoded literal regardless.

## Test plan

- [x] Unit tests added / updated
- [x] Manual testing performed
- [x] `npm test` passes — full suite green: 6070 passed, 0 failed
- [x] `npm run typecheck` passes

New `test/discover-subdomains-resilience.spec.ts` (11 cases) covers every acceptance criterion in the issue:

| Scenario | Expected |
|---|---|
| crt.sh fails | Certspotter used, subdomains returned |
| All sources fail, warm cache | stale result with age |
| All sources fail, cold cache | explicit `sourceUnavailable` error |
| Per-source outcome | `ct_source` log line per source |
| `force_refresh` + total outage | still serves stale, never empty |
| Outage after a good result | cache not clobbered |
| Genuinely empty result | not stored as last-known-good |
| Stale empty set | never rendered as confident "none found" |
| Budget trip before direct sources | serves stale, not an error |

**40/40 pass** across the three discover-subdomains specs (`-resilience`, base, `-deadline`); typecheck and lint clean.

One case deliberately exercises the handler, not just the tool: it guards the production seam where `scanCacheKV` is threaded in as `cacheKv`. Without that thread the resilience cache would silently never populate in prod while every unit test still passed.

### Suite status: green (6070 passed, 0 failed)

An earlier revision of this description reported 8 failures as unexplained pre-existing breakage. They are now fully diagnosed and resolved — **neither was a code defect.** Both were missing/stale *generated, gitignored* artifacts in a fresh worktree:

- **7 scoring failures** — a worktree has no `node_modules`, so `@blackveil/dns-checks` resolves via the parent checkout's symlink to the **parent's** `packages/dns-checks/dist/`. That dist predated the 3.36.0 composite-scoring fix (`938eccfa`), so tests written for the new behaviour ran against old compiled logic (`expected 62, received 85` — exactly the phantom-100 result). Proven with the repo's own paired wrappers: the SOURCE-side spec passed 22/22 while the BUILT-side spec failed 4 of the *same* shared assertions. `npm -w packages/dns-checks run build` in the parent fixes it.
- **1 wasm failure** — `crates/bv-wasm-core/pkg/` is gitignored build output, absent in a fresh worktree, so the suite collected 0 tests. `wasm-pack build --target web` fixes it.

Neither affects CI or deploys: `ci.yml` builds dns-checks (line 34) and WASM (line 59) before `npm test`, and both `build` and `deploy:prod` are prefixed with `npm -w packages/dns-checks run build`, so a deploy cannot ship a stale scoring bundle. Full detail, including two real defects that restoring the artifacts uncovered (a latency regression in this PR's own retry ordering, and a pre-existing mock leak in `free-tier-gating.spec.ts` that was making live network calls), is in the follow-up comment and commit `3f369393`.

Reviewers working in a worktree: use `npx vitest run`, not `npm test` — the wrapper resolves a worktree-local `node_modules/vitest/vitest.mjs` that doesn't exist there.

## Checklist

- [x] Changes follow existing code conventions
- [x] No secrets, credentials, or internal references included
- [x] Error messages follow the safe-prefix convention (see CLAUDE.md) — no new client-visible error strings
- [x] New tool? Followed the "Adding a New Tool" checklist in CLAUDE.md — n/a, no tool added or removed, `TOOL_DEFS` and the tool count are untouched

`discover_subdomains` is `scanIncluded: false` and not a scored category, so scoring surfaces, category counts, and score snapshots are unaffected. The added `stale` / `cacheAgeMinutes` fields are additive and optional on the existing custom-shape payload (the tool is already in `NON_CHECK_RESULT_TOOLS`).


MadaBurns:
### Update: the 8 "pre-existing failures" are fully diagnosed — suite is now green

The earlier note in this PR said 8 failures were pre-existing and deserved a separate issue. I investigated; **neither was a code defect, and the suite is now green at 6070 passed / 0 failed.** Two distinct causes, plus two real defects found along the way.

**Cause A — stale generated artifact (7 scoring failures).** `@blackveil/dns-checks` resolves through `node_modules/@blackveil/dns-checks`, a symlink to `packages/dns-checks`. A git worktree has no `node_modules`, so resolution walks up to the **parent checkout** — meaning tests import the *parent's* `dist/`, and building the package inside the worktree has no effect. The parent's `dist/` was built at 17:21 while the 3.36.0 composite-scoring fix (`938eccfa`, which removed the seed-to-100 loop) landed at 21:41, so tests written for the new behaviour ran against old compiled logic — hence `expected 62, received 85`, exactly the phantom-100 result.

Proven by the repo's own paired wrappers: `packages/dns-checks/src/__tests__/scoring/scoring-engine.spec.ts` (SOURCE) passed 22/22 while `test/scoring-engine.spec.ts` (BUILT) failed 4 of the *same* shared assertions from `scoring-engine.suite.ts`. Rebuilding the parent's dist (`npm -w packages/dns-checks run build`, per CLAUDE.md step 6) turned all 7 green; the bundle size changed 51094 → 50979 bytes, confirming the content genuinely differed.

**Cause B — missing generated artifact (1 wasm failure).** `crates/bv-wasm-core/pkg/` is gitignored build output and simply absent in a fresh worktree, so `test/wasm-integration.test.ts` collected 0 tests. `wasm-pack build --target web` in `crates/bv-wasm-core` fixed it (4/4).

**Not a CI risk, and not a deploy risk.** `ci.yml` explicitly builds dns-checks (line 34) and WASM (line 59) before `npm test`, and both `build` and `deploy:prod` are prefixed with `npm -w packages/dns-checks run build`, so a deploy cannot ship a stale scoring bundle. The gap is only that `npm test` alone has no prebuild step, so a local run in a worktree can silently test stale compiled logic. Worth knowing when working in a worktree; arguably worth a `pretest` hook, but that is out of scope here.

### Two real defects found and fixed (commit 3f369393)

Restoring the artifacts exposed a genuine failure that the stale build had been masking:

1. **A latency regression in this PR's own retry strategy.** Retrying a source that just timed out *before* trying the other source is the worst use of a tight budget — another full 10s timeout, while delaying the source that would actually rescue the caller. Worst case on the direct path was ~20.5s, which pushed `free-tier-gating.spec.ts` past its 15s timeout under load. `queryDirectSources` is now **failover-first, retry-second**: pass 1 tries every source once, and only a fully-failed pass spends remaining budget on a retry pass; backoff moved to between passes. Certspotter also gets a tighter 8s timeout (plain JSON API, unlike crt.sh's Postgres backend). Certspotter is now reached at ~10s instead of ~20.5s — the entire point of having a fallback. A new ordering test asserts the first pass is one attempt per source, not `['crtsh','crtsh']`.

2. **A pre-existing mock leak in `test/free-tier-gating.spec.ts`.** Its `afterEach` deliberately omitted `vi.resetModules()`, commented *"we don't mock modules across tests"* — but two tests do call `vi.doMock('../src/mcp/dispatch')`, and four earlier tests already import `../src/mcp/execute`. Since `doMock` only affects imports resolved after it, the cached graph holding the **real** dispatch won and those mocks silently no-op'd: the tool executed for real against unmocked fetch, leaving the test dependent on live CT-source wall-clock timing. The convention its own comment cites (`mcp-execute.spec.ts`) *does* reset modules. Adding `vi.restoreAllMocks()` + `vi.resetModules()` makes the mocks apply, so the test now asserts what it claims — and the file's runtime drops from **12.16s to 2.65s**, the removed time having been real network attempts.

The second one is the more valuable find: that test was passing for the wrong reason and making live network calls in CI.

PR #592, opened 2026-07-30
feat(scan-domain): emit individual findings in structuredContent

MadaBurns:
## Summary

- BizFit driver: a mobile app being built on `scan_domain` needs to show *which* findings, not just severity counts. `structuredContent.findingCounts` currently only tallies ("3 critical"); the full `Finding[]` those counts are filtered from already existed one line above in `buildStructuredScanResult` and was simply never emitted.
- Adds a `findings` array to `scan_domain`'s `structuredContent`, populated alongside `findingCounts`: same source (`result.score.findings`), same order (as produced, never re-sorted), no length cap, field-for-field passthrough of `category`/`title`/`severity`/`detail`. The internal `metadata` bag on `Finding` is deliberately excluded. `detail` is already sanitized by `createFinding()` at construction time and is passed through verbatim — no re-sanitization.
- **Additive and backwards-compatible**: every existing field of `StructuredScanResult` keeps its name, type, and value; `findingCounts` reports identical numbers before and after (asserted in the updated test).
- `scan_domain` is in `NON_CHECK_RESULT_TOOLS` (`src/schemas/tool-definitions.ts`), so it carries **no `outputSchema`** — confirmed, not assumed — and there is no parallel schema to update.
- `src/tools/batch-scan.ts`'s degraded-result placeholder (`emptyResult`, used for invalid-domain/budget-exceeded/thrown-scan batch entries) also gets `findings: []` to satisfy the now-required field on `BatchScanResultItem extends StructuredScanResult`.

## Test plan

- [x] `npx vitest run test/format-scan-report.spec.ts` — extended `buildStructuredScanResult returns machine-readable fields` to assert findings array length/order/round-trip, `metadata` exclusion, and that `findingCounts` is unchanged. 15/15 pass.
- [x] `npx vitest run test/format-scan-report.spec.ts test/handlers-tools.spec.ts test/scan-domain.spec.ts` — 143/143 pass.
- [x] `npm run typecheck` — clean.
- [x] `npm run lint` — clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ## [Unreleased]
 
+### Added
+
+- **`scan_domain`'s `structuredContent` now carries the individual findings, not just their counts.** `findingCounts` previously told a caller "3 critical" with no way to say which three — the full `Finding[]` the counts were already filtered from sat one line above, unemitted. A new `findings` array is added alongside `findingCounts`, additive and backwards-compatible: same source (`result.score.findings`), same order (as produced, never re-sorted), no length cap, field-for-field passthrough of `category`/`title`/`severity`/`detail` (the internal `metadata` bag is deliberately excluded). `detail` is already sanitized at `createFinding()` construction time and is passed through verbatim. `scan_domain` is in `NON_CHECK_RESULT_TOOLS` and so has no `outputSchema` to update for this change.
+
 ## [3.37.0] - 2026-07-29
 
 ### Added
--- a/src/tools/batch-scan.ts
+++ b/src/tools/batch-scan.ts
@@ -56,6 +56,7 @@ function emptyResult(domain: string, error: string): BatchScanResultItem {
 		maturityLabel: null,
 		categoryScores: {},
 		findingCounts: { critical: 0, high: 0, medium: 0, low: 0 },
+		findings: [],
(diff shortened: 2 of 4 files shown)


MadaBurns:
Confirmed and fixed in 334c7bc5 rather than dismissed — this was a real surface, not a theoretical one.

`check-dkim.ts:113-121` interpolates the `k=` value straight out of the scanned domain's own DKIM TXT record into the finding title (`Unknown DKIM key type: ${keyTypeMatch[1]}`, capture class `[^;\s]+`). Anyone can publish any TXT record on a domain they control, so before this fix a scan of an attacker's domain put attacker-authored text verbatim into every MCP client's `structuredContent`.

Worth noting the asymmetry you flagged is load-bearing in one direction: the prose path has always run titles through `sanitizeOutputText` (`format-report.ts:517,522`), and before this PR no machine-readable channel emitted titles at all. So this PR wasn't widening an existing raw-title surface — it would have opened a machine-readable bypass around a guard the human-readable channel already had.

Fixed at the emission site with the sanitizer this repo already trusts for titles, matching the adjacent `scoringSignals` treatment on the next line. Regression test is mutation-verified: reverting the call fails exactly one test.

Hardening `createFinding()` itself is the more durable fix — it would hold the guarantee at the one chokepoint instead of relying on every future emission site — but it changes a shared package's output for every consumer and every title, so I've left it as a follow-up rather than widening this PR's blast radius at merge time.

PR #610, opened 2026-08-02
fix(scoring): MTA-STS absence grades instead of zeroing; CAA TTL at the BR floor

MadaBurns:
## Summary

Two operator-approved scoring calibrations, both backed by a 1,000-domain local corpus scan, plus the mandatory `SCORING_MODEL_VERSION` bump.

### 1. MTA-STS absence no longer zeroes its category

`missingControl: true` zeroes a whole category. Measured over the corpus, the `mta_sts` category scored a **mean of 3.3 with 96.5% of the 687 measured domains at exactly 0**. A control almost nobody deploys separates nobody — it was acting as a flat constant penalty on ~3 of the ~80 base points rather than as a discriminator. Independent review also found no documented incident attributable to a missing MTA-STS policy, against a ~29.6% misconfiguration rate among adopters.

Absence is now a **graded** `medium` (ordinary -15 penalty, category 85), matching CAA, SVCB-HTTPS and TLS-RPT, which already decline to zero on absence.

Three absence sites dropped the metadata, all on the live package path:

| Site | Condition |
| --- | --- |
| `mta-sts-analysis.ts` `getMtaStsTxtFindings` | no `_mta-sts` TXT RRset |
| `mta-sts-analysis.ts` `finalizeMissingMtaStsRecordFinding` | same finding, detailed re-emit |
| `check-mta-sts.ts` both-missing summary | MX present, neither MTA-STS nor TLS-RPT |

All three are the *same absence condition*. **Deployed-but-broken is untouched** — bad `version:`, missing `mode:`, an uncovered MX set and an unfetchable policy file keep their confident `high`/`medium` findings, which never carried `missingControl`. A new locked test asserts that boundary.

No weight (MTA-STS stays 3, protective), tier, grade band, severity penalty or profile-detection rule changed.

### 2. CAA long-TTL threshold moved to the CA/Browser Forum 8-hour floor

The threshold was an arbitrary 86400s (24h) with a hand-picked buffer rationale. The corpus refutes it: across all 161 CAA-publishing domains, **135 RRsets observed, maximum TTL 21600s (6h)** — the finding fired 0 times and could not fire.

The principled line is the BR reuse-window floor, 28800s (8h): the rule is `max(TTL, 8h)`, so at or below the floor the TTL widens nothing, and strictly above it the TTL governs. It will still fire close to never on real corpora — that is a property of real-world CAA TTL practice, not a dead detector, and the measured basis is recorded in the constant's doc comment so nobody re-tunes it to a round number.

### 3. `SCORING_MODEL_VERSION` -> 1.6.0

Mandatory: change 1 raises scores for the large majority of domains.

## Measured impact

Re-scanned a deterministic random sample of the corpus with this build and diffed each score against the corpus baseline. Numbers are in the PR discussion / handoff report.

## Verification

| Gate | Result |
| --- | --- |
| `npm run typecheck` | exit 0 |
| `npx eslint` (8 touched files) | exit 0 |
| `npm -w packages/dns-checks test` | 618 passed / 32 files |
| `npm test` | **6965 passed, 0 failed**; exit 1 solely from the documented fresh-worktree `test/wasm-integration.test.ts` suite-**load** error (`crates/bv-wasm-core/pkg/` is built only by `npm run build:wasm`) |

## Snapshots touched

- `packages/dns-checks/src/parity-fixtures.ts` — MTA-STS fixture "no records, MX present" `0/missingControl:true` -> `85/false`. This IS the change.
- `test/check-mta-sts-non-apex.spec.ts` — two cases asserted `missingControl === true`; now assert the medium severity (the applicability signal they actually guard) plus `score === 85`.
- `test/check-caa-enforceability.spec.ts` — boundary cases re-pointed from the 24h threshold to the 8h floor; added a "fires at 12h" case that was silent before.

## Not merging

Opened for review only.


MadaBurns:
## Measured score impact

Method: deterministic random sample of 150 domains from the 1,000-domain corpus (seeded shuffle, only rows with a numeric baseline `score`), re-scanned with this branch's local build via `node dist/stdio.js` (same JSON-RPC handshake as `scripts/ci/dogfood-scan.mjs`), `force_refresh: true`, concurrency 10. Each new `score` diffed against the `score` recorded in the corpus file.

150 scanned, 0 harness errors. 1 domain (`hnjing.net`) returned an ungraded result (8+ checks errored on this run → `evidenceInsufficient`) and is excluded, leaving **149 comparable**.

### All 149 comparable domains

| Metric | Value |
| --- | ---: |
| mean delta | **+1.46** |
| median delta | **+2** |
| min delta | −23 |
| max delta | +9 |
| moved at all | 104 / 149 (69.8%) |
| moved up | 100 |
| moved down | 4 |
| flat | 45 |
| **NIST 6-band crossings** | **17 / 149 (11.4%)** |

Band transitions: C→B 10, D→C 3, B→A 1, F→D 1 (up) · C→F 1, D→F 1 (down).

`mta_sts` category transition across the sample: **0 → 85 on 103 domains**; 43 stayed n/a (not applicable / not measured); the 2 domains that already published MTA-STS (95, 100) were unchanged.

### Attribution — clean vs. noisy

The corpus baseline and this rescan are separate live measurements, so network-dependent checks (`ssl`, `http_security`, DNS) can flip between `completed` and `timeout`/`error` independently of any code change — and an inconclusive check is *excluded and renormalized* (scoring model 1.4.0), which moves the score on its own. Splitting the sample on "did every category's `checkStatus` match the baseline":

| Subset | n | mean | median | min | max | up | down | band crossings |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| **CLEAN** (identical checkStatuses — pure scoring-change signal) | 141 | **+1.73** | +2 | **0** | **+4** | 96 | **0** | 14 (9.9%) |
| NOISY (≥1 check changed measurement status) | 8 | −3.25 | 0 | −23 | +9 | 4 | 4 | 3 (37.5%) |

**In the clean subset the minimum delta is 0 and nothing moves down** — exactly what removing a zeroing rule should produce. Clean-subset band crossings: C→B 9, D→C 3, B→A 1, F→D 1 — all upward.

Restricted further to the 98 clean domains whose `mta_sts` actually went 0 → 85: mean **+2.49**, range 0..+4 (`mail_enabled` n=70 mean +2.41; `enterprise_mail` n=28 mean +2.68). That is the expected size — MTA-STS is 3 of ~80 base points, and the category goes 0 → 85, not 0 → 100.

### Every downward mover, investigated

All four are in the noisy subset. None is attributable to this change.

| Domain | Delta | Cause |
| --- | ---: | --- |
| `nic.xxx` | −23 | `ssl` was 100/`completed` in the baseline, `timeout` on this run → excluded as inconclusive. SSL is heavily weighted in `web_only`. `mta_sts` is n/a for it both times. |
| `smcaas6j.com` | −14 | 4 categories flipped to `error`/`timeout` (`dane_https`, `dnskey_strength`, `mx`, `ptr`) and `dkim` newly measured at 50. Its `mta_sts` actually went n/a → 90 (up). |
| `111.nhs.uk` | −3 | The one with no non-completed checks now: `http_security` was `timeout` (excluded) in the baseline and **completed at 60** on this run, so a previously-inconclusive category joined the denominator at 60. `mta_sts` n/a both times. |
| `metropolinet.co.il` | −2 | `mta_sts` 0 → 85 (up, from this change) but `ssl` 100 → `timeout` and `http_security` `error` → `timeout`; the SSL loss outweighs the MTA-STS gain. |

CAA: the long-TTL finding fired 0 times in the sample, as predicted — no observed CAA TTL exceeds the 8-hour floor.

Issue #572, opened 2026-07-27
core: add Mailjet + count unrecognized ESPs in spf-trust-surface catalog (parity follow-up to #566)


MadaBurns:
Follow-up to #566 / PR #570.

PR #570 fixed the SPF trust-surface under-count in the **worker layer** (`src/tools/check-spf.ts` post-processor + revived `src/tools/spf-trust-surface.ts`), so `check_spf`, `scan_domain`, `assess_spoofability`, `simulate_attack_paths`, and `validate_fix` now recognize Mailjet and count unrecognized broad ESP includes. The parity-locked **core** was intentionally left untouched to avoid a gated publish + re-vendor.

**Remaining gap:** the core catalog `packages/dns-checks/src/checks/spf-trust-surface.ts` `MULTI_TENANT_PLATFORMS` (lines ~27-49) still has **no Mailjet entry**, and the count (`spf-trust-surface.ts:125`, `${matchedPlatforms.length} shared platforms`) still counts only recognized platforms. Any DIRECT consumer of the vendored core under-counts — notably **bv-web-prod**, which calls core `checkSPF` directly (not the bv-mcp worker `checkSpf`), so its scan SPF surface still reports e.g. "2 shared platforms" for a 3-ESP record.

**Proposed core change:**
1. Add `['mailjet.com', { name: 'Mailjet', risk: 'Any Mailjet customer can send as your domain' }]` to `MULTI_TENANT_PLATFORMS` (registrable key matches `spf.mailjet.com` + sub-hosts via the `endsWith('.'+key)` rule).
2. (Optional, recommended) count unrecognized broad shared-sender includes toward the trust-surface total rather than only cataloged platforms — mirrors the heuristic added in the worker layer in PR #570. This changes count semantics corpus-wide, so it's an explicit call.

**Release impact (why this is separate from #566):** a core change requires a `@blackveil/dns-checks` version bump → npm publish → re-vendor into bv-web-prod (update `vendor/`, `VENDORED_VERSION`, parity corpus in `parity.contract.test.ts`) → bv-web-prod redeploy. Operator-gated cross-repo release.

Once merged into the core, the worker-layer post-processor from PR #570 can optionally be simplified/removed (it becomes redundant), but it is harmless to keep as defense-in-depth.

MadaBurns:
**Verification pass (2026-08-04, against `main` @ a23962f4) — claims confirmed, part 1 implemented on a branch.**

Verified:
[...]

Issue #566, opened 2026-07-27
check_spf: trust-surface finding + count omit unrecognized ESP includes (Mailjet missed; 3 includes reported as "2 shared platforms")


MadaBurns:
## Summary

`check_spf` under-reports the SPF trust surface when an `include:` points at an ESP its platform catalog doesn't recognize. For a record with **three** ESP includes, the findings named only two and the summary said **"2 shared platforms"** — omitting Mailjet. `resolve_spf_chain` and `map_supply_chain` both correctly enumerated all three in the same session, so the gap is specific to `check_spf`'s platform recognizer + its trust-surface count.

## Reproduction

Target: `spotto.ai`.

Actual SPF (verified via DoH):
```
v=spf1 include:spf.mailjet.com include:spf.protection.outlook.com include:441725607.spf01.hubspotemail.net -all
```

`check_spf spotto.ai` findings:
- "SPF delegates to shared platform: **Microsoft 365**"
- "SPF delegates to shared platform: **HubSpot**"
- "SPF trust surface: **2 shared platforms** (Microsoft 365, HubSpot)"

→ `include:spf.mailjet.com` (Mailjet — a multi-tenant ESP) is not recognized, so it is excluded from both the per-include findings and the platform count.

Cross-check, same session:
- `resolve_spf_chain` → 3 includes, all resolved (Mailjet, Outlook, HubSpot).
- `map_supply_chain` → lists `spf.mailjet.com` as an `email-sending` dependency (`trustLevel: critical`).

## Impact

The trust-surface finding is the one a reviewer reads to enumerate "who can send as this domain." Missing a recognized-in-other-tools ESP under-states the exposure and makes the count unreliable. Same class as the previously-closed catalog gaps #286 / #283 (which were on `map_supply_chain`), but on `check_spf`.

## Suggested fix

- Add Mailjet (`spf.mailjet.com`) to the shared-platform catalog.
- More robustly: when an `include:` resolves to a broad multi-tenant sender the catalog can't name, still count it toward the trust surface as an "unrecognized shared sender" rather than dropping it — the count should reflect the number of delegated includes, not only the number of *recognized* platforms.
- Ideally share one ESP/platform catalog across `check_spf`, `resolve_spf_chain`, and `map_supply_chain` so they can't disagree.

PR #614, opened 2026-08-04
fix(core): add Mailjet to the SPF trust-surface catalog (#572 part 1)

MadaBurns:
Closes the core half of #572 (part 1 only).

## What

The worker layer learned Mailjet in PR #570 (#566), but the parity-locked core catalog did not. DIRECT consumers of the vendored package — notably bv-web-prod, which calls core `checkSPF` rather than the bv-mcp worker wrapper — therefore still under-count the SPF trust surface, reporting e.g. "2 shared platforms" for a 3-ESP record.

- Adds the registrable key `mailjet.com` to `MULTI_TENANT_PLATFORMS`, so `spf.mailjet.com` and any Mailjet sub-host match via the existing `endsWith('.' + key)` rule.
- Adds `packages/dns-checks/src/__tests__/checks/spf-trust-surface.test.ts` — the core had no trust-surface coverage at all, so nothing tripped on the worker/core catalog drift.

## What this deliberately does NOT do

Part 2 of #572 (counting unrecognized broad shared senders toward the total) is **not** included. It moves the `matchedPlatforms.length > 1` emission threshold, so a "1 recognized + 1 unrecognized ESP" domain would start emitting a trust-surface finding corpus-wide and the SPF parity corpus would move. That is an operator call, not a config edit. Left as a documented `it.todo` rather than silently adopting either semantics.

## Release impact

This is a core change, so the bv-web-prod under-count does not clear until: `@blackveil/dns-checks` version bump (currently 1.10.0 — the exact version vendored) → npm publish → re-vendor into bv-web-prod (tarball + sha256 + `VENDORED_VERSION` in BOTH `shared/lib/scan/parity.contract.test.ts` and `shared/lib/mcp-scan/mcp-grade-parity.contract.test.ts`) → redeploy. Operator-gated cross-repo release; check the app Worker bundle budget before re-vendoring.

## Verification

- `packages/dns-checks` suite: 621 passed / 33 files, exit 0 (incl. `parity-corpus.contract.test.ts`)
- Rebuilt `dist` confirms Mailjet compiled into the published artifact
- Worker `spf-trust-surface` + `check-spf` + `scan-domain` specs: 87 passed, exit 0
- `npm run typecheck` and `npm run lint`: clean

### DIFF
--- a/packages/dns-checks/src/__tests__/checks/spf-trust-surface.test.ts
+++ b/packages/dns-checks/src/__tests__/checks/spf-trust-surface.test.ts
@@ -1,0 +1,68 @@
+// SPDX-License-Identifier: BUSL-1.1
+
+/**
+ * Core SPF trust-surface catalog coverage.
+ *
+ * The worker layer (`src/tools/spf-trust-surface.ts`) carries its own copy of this
+ * catalog, so a platform recognized there is NOT automatically recognized here —
+ * and DIRECT consumers of the vendored core (bv-web-prod calls `checkSPF` from
+ * the published package, not the bv-mcp worker wrapper) see only this file's
+ * catalog. Issue #572 was exactly that drift: Mailjet was added worker-side in
+ * PR #570 and left out of the core, so bv-web-prod under-counted the trust surface.
+ *
+ * These cases pin the core half of that parity.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { analyzeTrustSurface } from '../../checks/spf-trust-surface';
+
+describe('core SPF trust-surface catalog — Mailjet (#572)', () => {
+	it('recognizes the canonical spf.mailjet.com include', () => {
+		const findings = analyzeTrustSurface('v=spf1 include:spf.mailjet.com -all');
+
+		expect(findings).toHaveLength(1);
+		expect(findings[0].metadata?.platform).toBe('Mailjet');
+		expect(findings[0].metadata?.trustSurface).toBe(true);
+		expect(findings[0].metadata?.includeDomain).toBe('spf.mailjet.com');
+	});
+
+	it('matches Mailjet sub-hosts via the registrable-key suffix rule', () => {
+		const findings = analyzeTrustSurface('v=spf1 include:eu.spf.mailjet.com -all');
+
+		expect(findings).toHaveLength(1);
+		expect(findings[0].metadata?.platform).toBe('Mailjet');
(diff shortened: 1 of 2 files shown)


MadaBurns:
### Code review

Reviewed directly rather than via agent fan-out — the production change is a four-line catalog entry, and the one non-obvious risk (double-emission against the worker post-processor) was better answered by reading `augmentTrustSurface` and probing the analyzer than by parallel scans.

Found 1 issue:

1. **This change is not score-neutral on the worker path, contrary to how the PR body and #572 frame part 1.** The worker post-processor rebuilds its DMARC context from `coreTrustFindings[0]?.metadata` and falls back to `{}` when the core recognized nothing. Before this change a Mailjet-only SPF record produced zero core trust-surface findings, so the worker always defaulted to no-corroboration and emitted `info`. Now the core recognizes Mailjet and carries real `dmarcCorroborated` metadata, so the worker reconstructs the true context.

   Probed against the analyzer directly: `v=spf1 include:spf.mailjet.com -all` yields `['medium']` with `{corroboratedByWeakDmarc: true, dmarcPolicy: 'none', dmarcAlignmentMode: 'relaxed'}` and `['info']` with `{}`. A medium carries a -15 category penalty, so Mailjet-using domains with `p=none` + relaxed alignment will see their SPF category score drop.

https://github.com/MadaBurns/bv-mcp/blob/94e6859c26ca3fdac5aec04e015847380a6f6e96/src/tools/check-spf.ts#L69-L80

Assessment: this is correct behavior, not a defect — the corroboration rule already applies to the other 18 cataloged platforms, and Mailjet was the anomaly. But it is a scoring effect, it is currently unmeasured against the corpus, and it reaches production only at the version bump + publish, not at merge. Flagging it as a release-gate item rather than a merge blocker: the operator should decide whether to measure the Mailjet + weak-DMARC population before the release that ships this, and the same escalation lands on bv-web-prod at re-vendor.

No double-emission: `augmentTrustSurface` filters core trust-surface findings out and substitutes worker ones, so the now-redundant worker catalog entry does not stack with the core's.

Verified no other issues: catalog key is the registrable form (matches sub-hosts via the existing suffix rule, consistent with `_spf.salesforce.com`-style entries only in intent, not form — the catalog is already mixed on this), test file placement and `.test.ts` suffix match the sibling core check tests, and no scoring-model surface (`CATEGORY_TIERS`, weights, profiles) is touched.

Issue #417, opened 2026-06-19
M365 identity_secops: bv-mcp-side tracking for live Graph reads (upstream: bv-web-prod#2276, decision: #759)


MadaBurns:
Follow-up from #403 (M365 proxy repointed at bv-web-prod, shipped in bv-mcp PR #415 + bv-web-prod #467). The repoint cleared the `m365_proxy_404` by pointing the 4 `identity_secops` tools (`query_signins`, `query_ual`, `get_ca_policies`, `assess_coverage`) at a real bv-web-prod internal route — but that route currently returns a `representative: true` **seam**, not live Microsoft Graph data. Three follow-ups:

### 1. Live Microsoft Graph reads (feature)
bv-web-prod has a real M365 engine (`shared/lib/m365/graph-fetcher.ts`, `runComplianceScan`) but it's operator-gated (unbound `M365_DB` + Graph token) and shaped as a cron drift-sweep, not 4 per-`ms_tenant_id` query endpoints. Wiring true live reads (incl. the deferred `/auditLogs/signIns` feed) is new design: per-tenant Graph reads, `M365_DB` binding, token store. Tracked here.

### 2. Caller-facing `representative: true` labeling (MEDIUM, LLM-safety)
While the seam is in place, the 4 tools can return `{ ok: true, data: { representative: true, … } }` with `isError: false`. An LLM summarizing the result could present sample data as real threat intel. Add a note in the `TOOL_DEFS` descriptions (`src/schemas/tool-definitions.ts`) that `representative: true` in the body indicates sample/non-live data, until #1 lands.

### 3. Cross-repo contract test for the M365 internal seam (MEDIUM)
No contract test pins the bv-web-prod `/api/internal/mcp/m365/:tool` response envelope that bv-mcp's proxy classifies (`m365_proxy_<status>`, `unprovisioned`, `ok/data`). A future bv-web-prod envelope change would silently break bv-mcp's error classification. Mirror the `validate-key.contract.test.ts` pattern.

Refs: review of PR #415/#467; `src/tools/m365/proxy.ts`, bv-web-prod `app/routes/api/internal/mcp/m365.ts`.

MadaBurns:
**Status: 2 of 3 follow-ups SHIPPED; only the operator-gated live-Graph item remains. Keeping this open to track that one item.**

### ✅ #2 Caller-facing `representative: true` labeling — SHIPPED
[...]

PR #761, opened 2026-08-21
chore: release 3.61.1

MadaBurns:
Patch release — **tool description text only**, no behaviour, argument or response-shape change.

## What ships

**#417** — `query_signins`, `query_ual` and `assess_coverage` have no live Microsoft Graph path; they return sample data flagged `representative: true`. Their descriptions carried that caveat only *after* binding boilerplate, so a client listing tools saw the capability first and the caveat last. Each now opens with `SAMPLE DATA ONLY — …must NOT be used to investigate a real incident…`.

`get_ca_policies` is deliberately **excluded** — it has a live Entra path, so a blanket sample-only label would be the same untruth inverted. Both directions are pinned by contract tests (merged in #760).

## Why deploy

Verified against live prod at 3.61.0: all three tools still bury the disclosure. The mitigation only reduces real exposure once deployed.

## Version surfaces

| surface | value |
|---|---|
| package.json / lock / server.json | 3.61.1 |
| CHANGELOG heading | [3.61.1] |
| @blackveil/dns-checks | 1.23.0 (unchanged — zero files touched since v3.61.0) |

## Verification

- `build` 0 · `typecheck` 0 · `typecheck:tests` OK (no file exceeds baseline) · `lint` 0
- `scan-version-stamps` + `m365-tool-seam-wiring.contract` → 29 passed
- Full suite ran green on this content at #760 (659 files / 7781 tests)

No scoring-model change. `SCORING_MODEL_VERSION` unchanged; no domain is re-graded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/), and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [3.61.1] - 2026-08-21
+
+**No scoring-model change.** `SCORING_MODEL_VERSION` is unchanged — no check, weight, threshold or grade band moves, and no domain is re-graded. `@blackveil/dns-checks` stays at **1.23.0** (untouched this release). No tool behaviour, argument or response shape changes: this release alters tool *description* text only.
+
+### Changed
+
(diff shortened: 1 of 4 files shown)


MadaBurns:
Superseded by the 3.62.0 release. The #760 disclosure change is already on `main`; folding it into one release with the M365 client-tenant read kill switch (#762) avoids deploying an interim mitigation to production minutes before the capability it describes is disabled outright. Nothing is lost — this PR carried only version-surface bumps, no code.

PR #840, opened 2026-08-29
fix(ssl): inspect HSTS on HTTPS redirects

### LINKED ISSUE #839: check_ssl asserts "HSTS properly configured" (score 100, controlPresent:true) on any apex that redirects over HTTPS — the header is never read
## Summary

`check_ssl` emits its affirmative HSTS finding from the **absence of findings**, not from observing the header. For any apex that answers `https://` with a redirect to another HTTPS URL, the HSTS header is **never read at all** — so the check returns `passed: true`, `score: 100`, `controlPresent: true` and the text *"HSTS header is properly configured"* for a domain publishing no HSTS header.

This is a false **affirmative**, the dangerous direction, and it is the same class as #705 / #706 / #725 / #809 (`passed`-as-verdict) and #793 (`check_dnssec` false pass) — this time on `check_ssl`. It is the mirror image of #574, which was the false *negative* on the same check.

## Root cause

`packages/dns-checks/src/checks/check-ssl.ts:113-119`

```ts
const isRedirect  = response.status >= 300 && response.status < 400;
const location    = isRedirect ? response.headers.get('location') : null;
const isDowngrade = location?.startsWith('http://') ?? false;
const isHttpsRedirect = isRedirect && !isDowngrade;

if (!isHttpsRedirect) {
    findings.push(...getHttpsFindings(domain, redirectTarget, response.headers.get('strict-transport-security')));
}
```

`getHttpsFindings` is the **only** consumer of `strict-transport-security` in the package — one call site, confirmed by `grep -rn getHttpsFindings packages/dns-checks/src`. It is gated behind `!isHttpsRedirect`, so an apex that redirects HTTPS→HTTPS skips HSTS evaluation entirely.

Then, at `:55-64`:

```ts
if (findings.length === 0) {
    findings.push(createFinding('ssl', 'HTTPS and HSTS properly configured', 'info',
        `HTTPS connection succeeded and HSTS header is properly configured for ${domain}. …`));
}
```

Zero findings is treated as proof of a healthy control. It is not — here it is proof that nothing was measured.

**The affirmative is therefore never *earned* for any HTTPS-redirecting apex.** It is emitted regardless of the header, and happens to be true only when the header coincidentally exists. `blackveilsecurity.com` is in exactly that position: it also returns a 301 at the apex, so its `ssl: 100` comes from the same skipped branch — it is right by luck, not by measurement.

## Reproduction

Engine 1.25.0 / model 1.14.0 / cfg `9b752207`. Confirmed 2026-08-29 with `force_refresh: true`.

```
check_ssl { domain: "tauranga.govt.nz", force_refresh: true }
```

```json
{
  "category": "ssl",
  "passed": true,
  "score": 100,
  "controlPresent": true,
  "findings": [{
    "title": "HTTPS and HSTS properly configured",
    "severity": "info",
    "detail": "HTTPS connection succeeded and HSTS header is properly configured for tauranga.govt.nz. …",
    "metadata": { "confidence": "deterministic" }
  }]
}
```

Note `"confidence": "deterministic"` on an assertion that was never evaluated.

### The trigger condition

```
$ curl -sS -I https://tauranga.govt.nz/
HTTP/2 307
location: https://www.tauranga.govt.nz/
x-xss-protection: 1; mode=block
x-frame-options: SAMEORIGIN
```

A 307 to an HTTPS URL ⇒ `isHttpsRedirect === true` ⇒ HSTS never read.

### Two independent vantages agree the header is absent

**1. Google's `hstspreload.org`, fetching from its own infrastructure:**

```
$ curl -s "https://hstspreload.org/api/v2/preloadable?domain=tauranga.govt.nz"
{"errors":[
  {"code":"response.no_header","summary":"No HSTS header",
   "message":"Response error: No HSTS header is present on the response."},
  {"code":"redirects.http.www_first", …}]}
```

**2. Direct fetch, apex and www:**

```
$ curl -sSI -L https://tauranga.govt.nz/     | grep -i strict-transport-security   # (nothing)
$ curl -sSI -L https://www.tauranga.govt.nz/ | grep -i strict-transport-security   # (nothing)
```

An IP-pinned `--resolve` request and a full unfiltered header dump agree — the only security headers present are `x-xss-protection` and `x-frame-options`.

## Impact

Three wrong signals at once — `passed`, `score` and `controlPresent`. The `controlPresent: true` is the worst of them: it makes every downstream consumer that treats the flag as a presence oracle inherit the false affirmative, with no way to distinguish an asserted-present HSTS from a measured one. It also **suppresses a real missing control** from the customer's finding list.

Apex→www redirection is extremely common, so this is not an edge case. In an estate sweep of NZ government and health domains (2026-08-29) it affects a large share of the cohort. Any report quoting `ssl: 100` as "HSTS is fine" is unsafe.

## Suggested fix

Read the header on the response that was actually fetched, independent of the redirect branch, and never emit an affirmative from an empty finding list:

1. Evaluate `strict-transport-security` on **every** assessable response, including HTTPS redirects. An HSTS header on the redirect response is valid and common (it is how a preloaded apex works).
2. If the header is absent → emit the missing-HSTS finding, `controlPresent: false`, score below 100.
3. If the posture genuinely could not be measured → `checkStatus: 'error'` and exclude the category, as the `robotsDisallowed` and `inconclusive` paths already do correctly at `:36` and `:44`.
4. Replace the `findings.length === 0` affirmative with one derived from a positive observation. "No findings" must never be the evidence for a control being present — that is the defect shape shared with #705/#706/#793.

## Regression test

`packages/dns-checks/src/__tests__/checks/check-remaining.test.ts:234` currently asserts the affirmative title is present; it will need to change with the fix.

Add a case: apex returns **301/307 to an HTTPS URL with no `strict-transport-security` header** → assert `controlPresent === false`, `score < 100`, and that **no finding title or detail asserts HSTS is configured**. Assert on the detail string specifically — score and prose have drifted apart here, and only the prose is customer-visible.

Filed from bv-web-prod; found during an NZ government/health estate sweep.

mikemikimike:
## Issue

Fixes #839.

## Background

`check_ssl` skipped HSTS analysis whenever the HTTPS probe returned an HTTPS-to-HTTPS redirect, causing an empty finding set to be reported as a successful HSTS configuration.

## Changes

- Analyze the HSTS header on HTTPS redirect responses.
- Add a regression test for an HTTPS redirect without HSTS.

## Compatibility

Existing non-redirect and downgrade behavior is preserved. HTTPS redirects without HSTS now report the missing control instead of a false affirmative.

## Tests

- `npm test -- packages/dns-checks/src/__tests__/checks/check-remaining.test.ts` — blocked by the local Cloudflare Workers pool failing to spawn (`EFTYPE`); no tests started.
- `npx prettier --check packages/dns-checks/src/checks/check-ssl.ts packages/dns-checks/src/__tests__/checks/check-remaining.test.ts` — initially found the new test formatting; formatted only the touched test file.
- `git diff --check` — passed.

## Limitations

The focused Vitest suite, full test suite, typecheck, lint, and build could not be completed because the required Cloudflare Workers test runtime failed to spawn in this Windows environment.

### DIFF
--- a/packages/dns-checks/src/__tests__/checks/check-remaining.test.ts
+++ b/packages/dns-checks/src/__tests__/checks/check-remaining.test.ts
@@ -73,11 +73,7 @@ describe('checkCAA', () => {
 
 	it('returns info for properly configured CAA', async () => {
 		const queryDNS = createMockDNS({
-			'example.com': [
-				'0 issue "letsencrypt.org"',
-				'0 issuewild "letsencrypt.org"',
-				'0 iodef "mailto:security@example.com"',
-			],
+			'example.com': ['0 issue "letsencrypt.org"', '0 issuewild "letsencrypt.org"', '0 iodef "mailto:security@example.com"'],
 		});
 		const result = await checkCAA('example.com', queryDNS);
 		expect(result.findings.some((f) => f.title === 'CAA properly configured')).toBe(true);
@@ -162,7 +158,9 @@ describe('checkNS', () => {
 		// A thrown NS query is a transient resolver failure — we could not MEASURE the posture, so
 		// the category is excluded from scoring (checkStatus 'error') with an info-only finding,
 		// rather than penalizing a possibly-healthy domain with a critical "NS query failed".
-		const queryDNS: DNSQueryFunction = vi.fn(async () => { throw new Error('fail'); });
+		const queryDNS: DNSQueryFunction = vi.fn(async () => {
+			throw new Error('fail');
+		});
 		const result = await checkNS('example.com', queryDNS);
 		expect(result.checkStatus).toBe('error');
 		expect(result.findings[0].severity).toBe('info');
@@ -227,7 +225,7 @@ describe('checkSSL', () => {
 			// HTTP redirect to HTTPS
 			return new Response('', {
 				status: 301,
-				headers: { 'location': 'https://example.com/' },
+				headers: { location: 'https://example.com/' },
 			});
 		});
 		const result = await checkSSL('example.com', fetchFn);
@@ -235,9 +233,26 @@ describe('checkSSL', () => {
 	});
 
 	it('reports connection failure', async () => {
-		const fetchFn: FetchFunction = vi.fn(async () => { throw new Error('Connection failed'); });
+		const fetchFn: FetchFunction = vi.fn(async () => {
+			throw new Error('Connection failed');
+		});
 		const result = await checkSSL('example.com', fetchFn);
 		expect(result.findings.some((f) => f.title === 'HTTPS connection failed')).toBe(true);
+	});
+
+	it('reports missing HSTS when HTTPS redirects to another HTTPS URL', async () => {
+		const fetchFn: FetchFunction = vi.fn(async (url: string) => {
+			if (url === 'https://example.com') {
+				return new Response('', { status: 307, headers: { location: 'https://www.example.com/' } });
+			}
+			return new Response('', { status: 200, headers: { 'strict-transport-security': 'max-age=31536000; includeSubDomains' } });
+		});
+
+		const result = await checkSSL('example.com', fetchFn);
+
+		expect(result.findings.some((f) => f.title === 'No HSTS header')).toBe(true);
+		expect(result.findings.some((f) => f.title === 'HTTPS and HSTS properly configured')).toBe(false);
+		expect(result.score).toBeLessThan(100);
 	});
 });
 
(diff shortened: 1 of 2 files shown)


MadaBurns:
### Code review

Found 1 issue:

1. Removing the `isHttpsRedirect` guard makes `getHttpsFindings()` read the HSTS header off the intermediate 3xx response itself (the probe uses `redirect: 'manual'`), not the final destination. Sites that set HSTS only on the terminal 200 — common with CDN/edge redirects — will now take a scored "No HSTS header" medium finding despite correct HSTS, trading the old false pass for a new false penalty. The guard was added deliberately in 642cbb3c ("skip HSTS check on HTTPS-to-HTTPS redirects where we can't see the final page headers"); its explanatory comment was lost in the monorepo port, so this wasn't visible in the file. The sibling check documents the same hazard: `followRedirects()` in `check-http-security.ts` exists because redirect responses "typically lack security headers, causing false negatives if we analyze the 301 instead of the 200". The new regression test encodes the problem as intended behavior — the redirect target `https://www.example.com/` carries valid HSTS, yet the test asserts "No HSTS header":

https://github.com/MadaBurns/bv-mcp/blob/8ffb24959a9edce55aba35a1e3e024079cb5db19/packages/dns-checks/src/checks/check-ssl.ts#L133-L138

https://github.com/MadaBurns/bv-mcp/blob/8ffb24959a9edce55aba35a1e3e024079cb5db19/packages/dns-checks/src/__tests__/checks/check-remaining.test.ts#L242-L248

Suggested direction: keep the unconditional HSTS inspection only if the redirect is followed to the final response first (mirroring `followRedirects()` in `check-http-security.ts`), or — if scoring the first HTTPS response is the intended standard (it matches the hstspreload.org requirement that HSTS be present on the initial redirect response) — state that rationale in an in-code comment so the March fix isn't re-litigated a third time. Note also that `const isHttpsRedirect = ...` is now unused and will fail the repo's `no-unused-vars` lint in the required `build-and-test` check.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MadaBurns:
Thanks @mikemikimike — the false-pass find in #839 was real and your fix was the right instinct. Review turned up a wrinkle: the removed guard was added deliberately in 642cbb3c because the hop-only header read penalizes sites that set HSTS on the terminal response (the comment explaining that was lost in the monorepo port, so there was no way to see it). #843 builds on your commit (retained for attribution) and extends it to follow the redirect chain, with the unresolvable case routed to the inconclusive lane. Appreciated!

Issue #832, opened 2026-08-28
check_lookalikes: throttled run publishes a CONTRARY ownership attribution instead of unmeasured (jcpenny.com: third_party at 63/94 unresolved, owned_by_seed at 1/94)


MadaBurns:
## Summary

Under throttling, `check_lookalikes` publishes a **contrary ownership attribution** instead of declining to attribute. A run with 63 of 94 candidate lookups unresolved reported `jcpenny.com` as `third_party` with the detail "no ownership signal links it to this organisation" — the signals weren't absent, they were **unfetched**. Near-complete runs of the same scan report `owned_by_seed` on a full 8/8 nameserver match.

## Measurement (2026-08-29, three sequential `force_refresh:true` runs on `jcpenney.com`)

| run | unresolved | complete | jcpenny.com verdict | detail (verbatim) |
| --- | --- | --- | --- | --- |
| 1 | **63 / 94** | false | `third_party` (2 findings incl. "Impersonation-shaped web infrastructure observed") | "jcpenny.com is registered with its own nameservers, distinct from jcpenney.com — no ownership signal links it to this organisation." |
| 2 | 1 / 94 | false | `owned_by_seed` | "The domain jcpenny.com is owned by the same organisation as jcpenney.com (jcpenny.com matches the complete 8/8 nameserver set on a shared provider. A full match is evidence; a partial match on the same provider would not be.)" |
| 3 | 1 / 94 | false | `owned_by_seed` | same as run 2 |

Run 1's own `scan_status` correctly reported the degradation — but the attribution layer still rendered a definitive negative ("no ownership signal") from data it had failed to fetch, and escalated the same domain into an impersonation-shaped finding against what runs 2–3 establish is the brand's own defensive registration.

## Why it matters

This is the non-answer-becomes-record defect class: a throttled lookup must surface as `unmeasured`, never as the **opposite** attribution. A consumer diffing runs sees the world flip (owned → third-party impersonation → owned) when only the instrument's completeness changed. It is the attribution-layer sibling of #781 (candidate-set non-determinism from silently dropped failed lookups, closed 2026-08-24): the drop is no longer silent in enumeration metadata, but its consequence still reaches the verdict.

## Suggested direction

[...]

PR #853, opened 2026-08-30
fix(lookalikes): stop the seed NS probe racing its own permutation fan-out

MadaBurns:
## The defect

The seed domain's NS lookup gates **every** ownership verdict in a lookalike run: when it fails, all candidates degrade to `unmeasured` and every impersonation-shaped finding is withheld (#832).

It was issued with `PHASE1_DNS_OPTS` (`timeoutMs: 2000, retries: 0`) inside the **same `Promise.all`** as `filterByNsExistence`, which dispatches ~66 candidate probes in adaptive batches. So the one query that could least afford to lose was competing with its own burst for resolver budget, with no retry to survive losing.

The lean preset is correct for the 66 candidates — a dropped probe costs one disposable permutation. It is wrong for the seed, where a drop voids the interpretation of the whole result. Same options object, opposite blast radius.

## How it was narrowed

| Hypothesis | Disconfirming check | Result |
|---|---|---|
| Real resolver outage | `dig NS` + raw DoH for `meta.com` | ✗ 4 answers, HTTP 200 |
| Rate-limited by external load | Re-run idle, zero concurrent load | ✗ failed again, 2/2 |
| Self-contention | Read the call site | ✓ confirmed |

The tell was the asymmetry: candidate failures *varied* run to run (37 → 32) while the seed failed **2/2**. Random throttling does not do that.

## The change

- Hoist the seed's NS + MX into a Phase 0 that completes **before** the fan-out starts.
- Give them `SEED_DNS_OPTS` (`timeoutMs: 5000, retries: 2`).

Costs one DNS round trip on a check that already takes seconds.

## Measured against the live resolver (`meta.com`)

| | before | after |
|---|---|---|
| `seedNsUnresolved` | **true** (2/2 runs) | **gone** |
| candidates resolved | 24–29 | **96** |
| `unmeasured` verdicts | all | **0** |
| attributed verdicts | 0 | **96** |

Candidate yield nearly quadrupled — the fan-out had been starving itself too.

## Scope note

Attribution now returns `third_party`, which is a **measurement result, not a claim of impersonation**. Legitimate unrelated businesses are third-party registered too. This PR restores the ability to attribute; it does not change what an attribution means.

## Verification

- Adds the **first test covering `seedNsUnresolved`** — it was previously untested.
- Full lookalike suite **96/96 green** (`check-lookalikes`, `lookalike-analysis`, `lookalike-severity`).
- Typecheck error set **byte-identical** to a clean `origin/main` baseline worktree: 409 pre-existing (missing `cf-typegen` CF globals), **0 in the changed files**.

### DIFF
--- a/src/tools/check-lookalikes.ts
+++ b/src/tools/check-lookalikes.ts
@@ -188,15 +188,19 @@ async function checkLookalikesCore(
 
 	const permsToProbe = [...nonDotInsertionPerms, ...filteredDotInsertionPerms];
 
+	// Phase 0 (#850): resolve the SEED's own NS + MX BEFORE the permutation
+	// fan-out, not alongside it. These two queries used to share a Promise.all
+	// with `filterByNsExistence`, which dispatches ~66 probes in adaptive
+	// batches — so the one lookup that gates every ownership verdict competed
+	// with its own burst for resolver budget, and (on PHASE1_DNS_OPTS) had no
+	// retry to survive losing. Measured 2026-08-31: meta.com reported
+	// `seedNsUnresolved: true` on 2/2 idle runs while a standalone DoH NS query
+	// for it returned 4 answers. Sequencing costs one round trip on a check
+	// that already takes seconds; a voided attribution costs the whole result.
+	const [primaryNsProbe, primaryMx] = await Promise.all([queryPrimaryNs(domain), queryPrimaryMx(domain)]);
+
 	// Phase 1: Fast NS existence check — filter out unregistered domains.
-	// Also query the primary domain's NS + MX for ownership comparison: NS for
-	// the ownership verdict itself, MX for the D4 MX-overlap corroboration
-	// signal consulted by `attributionConfidence()` (wording only).
-	const [nsResult, primaryNsProbe, primaryMx] = await Promise.all([
-		filterByNsExistence(permsToProbe),
-		queryPrimaryNs(domain),
-		queryPrimaryMx(domain),
-	]);
+	const nsResult = await filterByNsExistence(permsToProbe);
 	const { registered: registeredPerms, nsMap: lookalikeNsMap, unresolved: nsUnresolved } = nsResult;
 	const primaryNs = primaryNsProbe.ns;
 	// #832 — the seed's own NS lookup failed, so the ownership comparison has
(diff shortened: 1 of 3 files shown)


MadaBurns:
Merged. The production change is sound — sequencing the seed probe ahead of the fan-out and giving it its own retry preset is well-justified, and `contract` failed only on `test/oauth/register-rate-limit.spec.ts` (expected 429, got 201), which touches none of these three files and passed in `build-and-test` on the same commit. Flake, not a regression.

Two defects worth a follow-up, neither affecting shipped behaviour:

**1. `tset.com` is not a generated permutation of `test.com`, so two of the new test's three assertions are vacuous.**

I verified this against all three generators on `origin/main`:

- `generateLookalikes` has six rules — QWERTY-adjacent key swap, character omission, character duplication, dot insertion, TLD swap, homoglyph. **None is a transposition of the domain's own letters**, which is what `test` → `tset` requires.
- `generateCognitiveLookalikes` is morpheme swaps + cognitive substitutions + orthographic insertions. No transposition.
- `generateCombosquats` is brand + hyphenated affix. No.

So the `tset.com` mock routes are never requested, `registeredPerms` is empty, and the run takes the `buildNoRegisteredCandidatesFinding` path. `expect(unmeasured).toHaveLength(0)` therefore passes because there are **no ownership verdicts at all**, not because none was degraded — and the `seedNsUnresolved` assertion is likely vacuous for the same reason.

`expect(seedNsAttempts).toBeGreaterThan(1)` is real and does prove the retry, so the test isn't worthless — it just doesn't test the candidate-attribution claim its comment makes.

**`tes.com` is a generated permutation** (character omission at index 3) and is the substitution I landed on for exactly this reason in the #847 follow-up, after debugging it by surfacing the probed-candidate list through a deliberate assertion failure — `console.log` does not surface from the Workers pool.

**2. The `#850` references point at the wrong thing.**

`#850` is `fix(dnssec): a failed DS/DNSKEY lookup no longer caps the whole domain at grade D`, unrelated to lookalikes. The reference appears in the `Phase 0 (#850)` comment in `check-lookalikes.ts`, the `SEED_DNS_OPTS` JSDoc in `lookalike-dns.ts`, and the new test's name and comment — so a reader following any of them lands on a DNSSEC change. Worth repointing at the real issue, or dropping the number if the measurement in the comment is the whole provenance.

MadaBurns:
Filed as #858 so this does not live only as a comment on a merged PR.

PR #879, opened 2026-09-01
fix(deploy): close the injector's silent-drop gap, gate prod on required secrets, add a staged rollout

MadaBurns:
Aligns the deploy path with Cloudflare's current Workers deployment guidance. Three changes, all additive — `deploy:prod` is untouched and remains the authoritative path.

## 1. The injector was silently dropping private-overlay keys

`scripts/inject-private-config.cjs` merges `.dev/wrangler.deploy.jsonc` onto the public config through an **allowlist** of seven keys. Everything else in the overlay is discarded without a word — the exact "silent drop" class the deployment mandate in `CLAUDE.md` warns about.

The live overlay carried eight keys outside that allowlist, two of them disagreeing with the public config:

| Overlay key | vs public | Effect today |
|---|---|---|
| `compatibility_date` | 3 months stale | Discard is **load-bearing** — merging it would regress the runtime |
| `observability` | omits `traces` | Discard is correct |
| `durable_objects`, `migrations`, `name`, `compatibility_flags`, `upload_source_maps` | identical | Dormant — a future edit would vanish |
| `$schema`, `main` | relative to `.dev/` | Correct to discard |

Production was correct, but by coincidence rather than by design. Every overlay key is now classified: unknown → FATAL, public-owned drift → warning, `name`/`durable_objects`/`migrations` drift → FATAL (silently dropping a DO binding or a migration tag ships a config nobody reviewed).

## 2. `secrets.required`, scoped to the generated production config

Wrangler's `secrets` property makes `wrangler deploy` refuse to ship when a declared secret is not configured on the Worker.

Deliberately **not** added to the public `wrangler.jsonc`: declaring it there also makes `wrangler dev` load *only* the listed keys from `.dev.vars` and stops `wrangler types` inferring from it. Local `.dev.vars` holds ten keys, five of which are release tooling rather than Worker secrets, and several production secrets are absent from it — so a public declaration would warn on every `wrangler dev` and change behaviour for OSS self-hosters. Emitting it into the generated config gets the gate with neither side effect.

Each declared name was verified provisioned on the live Worker before being added, so **the gate cannot block the next deploy**. Fail-soft capabilities (`BV_RECON_KEY`, `BV_TLS_PROBE_KEY`, `CERTSPOTTER_TOKEN`, `BV_BROWSER_RENDERER_KEY`, `BV_CERTSTREAM_ADMIN_KEY`, the dev/mobile keys) are excluded — each is designed to degrade when unset, so requiring them would fail a deploy over a supported configuration.

## 3. Staged rollout path

`deploy:prod` ends in `wrangler deploy` — 100% of traffic to a brand-new version with no intermediate check. Adds `deploy:prod:staged` (identical gates, ends in `versions upload`, no traffic), plus `deploy:prod:promote` and `deploy:prod:triggers`.

The separate triggers step is required, not optional: `versions upload` does not apply cron or route changes, so a staged rollout would otherwise silently drop them. Note that preview URLs are not generated for Workers with Durable Objects, so smoke-testing a staged version goes through a version override rather than a preview URL.

Also adds `deploy:infra-probe` (`deploy:prod` does not deploy bv-infra-probe — previously only a `CLAUDE.md` footnote) and `dev:multi` for the multi-Worker local dev pattern, since the two Workers are joined by a service binding.

## Verification

- Injector output is **byte-identical** to the pre-change output when run against the real private overlay; the only difference is the two drift warnings. No deploy behaviour changes.
- `npm run check:bindings:prod` — the exact gate `deploy:prod` runs — passes with the `secrets` block present, confirming no `Env` type collision.
- `deploy-pipeline.audit.test.ts` 10/10, `private-config-injection.node.test.ts` 7/7, `infra-probe-wrangler` + `wrangler-public-no-private-bindings` audits green.
- eslint clean; `typecheck:tests` ratchet -4 (no regression).

New audits pin both paths: the staged path must carry every gate `deploy:prod` carries and must never contain `wrangler deploy`; the injector must declare the load-bearing secrets and must not declare the fail-soft ones.

## Follow-up, not in this PR

`scripts/deploy-private.mjs` runs `wrangler deploy --config .dev/wrangler.deploy.jsonc` **directly**, bypassing the injector and therefore every gate above plus `validateProductionSecurityConfig`. It is wired to no npm script and reads like the BSL self-hoster path (pairs with `wrangler.private.example.jsonc`), where bypassing is correct — but run against the real overlay it deploys ungated. Needs a decision: route it through the injector, or rename it so its audience is unambiguous.


MadaBurns:
## Follow-up resolved: `deploy-private.mjs` now routes through the injector

Picked the "route through the injector" option from the follow-up note above, and it turned up a live defect rather than just a hygiene gap.

**The second door was shipping an incomplete Worker.** `wrangler.private.example.jsonc` — the template the script's own error message tells you to copy — carried its own `durable_objects` and `migrations` copies, and both had gone stale: one DO binding and one migration tag, against the public config's two and three. Deploying a fresh overlay through this door shipped **without `PROFILE_ACCUMULATOR`** (adaptive weights silently falling back to static) and with **no cron triggers at all**, because `triggers` only exists in the public config the door never read.

**Changes**

- `scripts/deploy-private.mjs` injects first and hands wrangler the generated `wrangler.production.jsonc`, so both doors deploy the same reviewed artefact. Step handling is factored into one helper that aborts on any non-zero exit.
- `wrangler.private.example.jsonc` is rewritten as what it actually is — an overlay, not a standalone config. It carries only the keys the injector merges and no longer repeats keys `wrangler.jsonc` owns, which is what let them rot unnoticed. It gains the four vars the injector requires, so copying it now produces a config that injects cleanly instead of failing on first deploy.

**Verification** — injecting the shipped example against the real public config now yields `QUOTA_COORDINATOR, PROFILE_ACCUMULATOR`, migrations `v1, v2, v3`, all three cron triggers, and `compatibility_date 2026-07-29` (the example previously pinned `2026-03-01`). Pinned by two new audits: the helper must inject before wrangler and must never pass the raw overlay to `deploy`; the shipped example must inject into a config that still has both DO bindings, all three migrations, and the crons.

**Scope note** — this door is now *config*-safe, not *release*-safe. It still does not rebuild `@blackveil/dns-checks` (the F2 gate), nor run `check:deploy-freshness`, `check:release-integrity`, or `check:bindings:prod`. `npm run deploy:prod` remains the authoritative path; closing that remaining delta is a separate call.

Issue #867, opened 2026-08-31
INVERSION of #780: registrationDays is now null for most .com candidates and populated for .ai — disabling the age filter where it is needed most


MadaBurns:
## Summary

#780 ("registrationDays is null for every `.ai` candidate, populated for `.org`/`.com`") was closed **COMPLETED 2026-08-24**. One week later the pattern has **inverted**: `.ai` candidates are populated, and the majority of `.com` candidates are null.

Measured 2026-08-31, engine 1.29.0 / model 1.18.0 / config `9b752207`.

## Evidence (first-hand, from the raw payloads)

| Seed | TLD | populated | null |
|---|---|---|---|
| `anthropic.com` | .com | **8** | **24** |
| `openai.com` | .com | 4 of 8 mail-capable | 4 of 8 |
| `x.ai` | .ai | **7** | 0 |
| `mistral.ai` | .ai | **4** | 0 |

Null `.com` candidates on the `anthropic.com` run include `anthropic.co`, `anthropic.org`, `anthropic.net`, `anthropic.io`, `anthrapic.com`, `anthropc.com`, `anthropi.com`, `anthropic-support.com`.

## Why this is worse than #780

`registrationDays` is the only field that supports the **age filter**, and the age filter is what removes a whole class of false positive: *a typosquat cannot predate the brand it targets*. In this sweep it correctly excluded:

- `googie.com` — ~1998-01-21, predates Google's founding
- `chere.com` — ~1999-02-03, predates Cohere (2019) by ~20 years
- `mitral.ai` — ~2021-01, predates Mistral AI (Apr 2023); "mitral" is also a cardiology term
- `mystral.ai` — ~2023-01, predates Mistral AI

With the field null, that filter cannot run. On `cohere.com`, **14 of 15** mail-capable candidates had null ages, so 14 domains could be neither confirmed nor excluded — reported as impersonation-capable purely because nothing could rule them out.

So the regression lands precisely on `.com`, where brand-adjacent squatting is most common, and it silently converts "unknown age" into "not excluded", which downstream reads as "confirmed".

## Suggested fix

1. Restore RDAP/WHOIS age lookup for `.com` (the `.ai` path evidently works now — whatever changed there should generalise).
2. Where age genuinely cannot be resolved, mark the candidate `ageUnknown: true` rather than leaving `registrationDays: null` to be read as absence, and **exclude it from the rollup count** rather than including it by default. Unknown is not a threat observation.

## Related

[...]

Issue #781, opened 2026-08-24
check_lookalikes: candidate set is non-deterministic across force_refresh runs, yet labelled confidence: deterministic


MadaBurns:
## Summary

Repeated `check_lookalikes` calls on the same domain with `force_refresh: true` return **different candidate sets** — differing in both count and membership — with no field indicating truncation, timeout, or partial enumeration. A consumer cannot distinguish "the set shrank because enumeration was incomplete" from "the set shrank because a domain was deregistered."

## Evidence — three runs against `openclaw.ai`, 2026-08-24, all `force_refresh: true`

| run | candidates | notable membership changes |
|---|---|---|
| 1 | 12 | baseline |
| 2 | 10 | `openklaw.ai`, `openslaw.ai` absent |
| 3 | 13 | `openklaw.ai`, `openslaw.ai` back; `opemclaw.ai`, `openflaw.ai`, `openclow.ai` appear for the first time |

Three names present in run 3 appeared in neither earlier run. All three runs were minutes to hours apart; no domain was actually registered or deregistered in that window.

Every finding in every run carries `"confidence": "deterministic"`.

## Why it matters

1. **`confidence: "deterministic"` is inaccurate** for a result set that is not reproducible. That label invites consumers to treat the output as a census.
2. **Longitudinal/monitoring use is unsound.** Anything diffing consecutive runs to alert on "new lookalike registered" will emit false alerts on names that were merely missed last time, and miss real ones that happened to be enumerated before.
3. **Reports overstate completeness.** A client-facing document naturally reads "fifteen confusable domains are registered" as the full set. It is a sample. Nothing in the response says so.

## Suggested fix

- Emit enumeration metadata: how many permutations were generated, how many were resolved, how many lookups timed out or were skipped, and whether any budget/deadline was hit.
- Downgrade `confidence` from `deterministic` when the run was not exhaustive.
- If the variance is caused by per-candidate DNS timeouts under a wall-clock budget, consider returning candidates in a stable order with an explicit `resolutionStatus` per candidate rather than dropping unresolved ones silently.

MadaBurns:
Fixed in `0638d28d` (#784).

PR #892, opened 2026-09-03
fix(lookalikes): throttled runs abstain from the threat rollup and report counts as floors (#865)

### LINKED ISSUE #865: REGRESSION of #832: a throttled check_lookalikes run still publishes a hard mail-capable count at HIGH severity instead of abstaining (openai.com, 77/90 unresolved)
## Summary

#832 ("throttled run publishes a CONTRARY ownership attribution instead of unmeasured", `jcpenny.com` at 63/94 unresolved) was closed **COMPLETED 2026-08-30**. One day later the adjacent half of that defect still reproduces: a run that resolved **13% of its permutation space** still emits a rollup with hard counts at `high` severity.

Measured 2026-08-31, engine 1.29.0 / model 1.18.0 / config `9b752207`, `force_refresh: true`. Reproduced twice, hours apart.

## Reproduction

`check_lookalikes` on `openai.com`:

```json
"enumeration": {
  "permutationsGenerated": 90, "permutationsProbed": 90,
  "candidatesResolved": 12, "unresolvedCount": 77, "complete": false
}
```

Alongside that, at severity `high`:

> **"6 lookalike domains showing pre-phishing staging signals"**
> `mailCapableCount: 8`, and prose: *"8 can send mail in total."*

The scan_status finding *does* say the list is a sample. But the threat rollup states `8` as a fact, in the same payload, at higher severity — and a consumer reading severity-ordered findings sees the hard number first.

## Why this is the same class as #832

#832 was about a throttled run publishing a *contrary attribution*. This is a throttled run publishing a *contrary confidence*: 85.6% blind, yet `confidence: "heuristic"` with an exact integer and no abstention. In both cases throttling changes the answer while the output does not say the answer is unreliable in the field a consumer keys on.

## Cross-provider impact — counts are not comparable

Same tool, same day, seven providers. The unresolved rate varies 1.2% -> 85.6%:

| Domain | probed | resolved | unresolved | mail-capable emitted |
|---|---|---|---|---|
| amazon.com | 82 | 80 | 1 (1.2%) | 6 |
| anthropic.com | 95 | 33 | 3 | 11 |
| perplexity.ai | 94 | 24 | 37 (39%) | 7 |
| cohere.com | 94 | 29 | 40 (43%) | 15 |
| microsoft.com | 95 | 42 | 48 (51%) | 10 |
| meta.com | 66 | 20 | 41 (62%) | 16 |
| google.com | 89 | 29 | 60 (67%) | 3 |
| openai.com | 90 | 12 | 77 (86%) | 8 |

Every one is `complete: false`. Reading these as a comparison implies OpenAI (8) is less targeted than Cohere (15), when the difference is dominated by how much of each namespace was reachable. **Publishing them side by side would report a rate-limiting artifact as a security difference about named companies.**

## Suggested fix

1. Above an unresolved-ratio threshold, the threat rollup must abstain — not-assessed, no integer, no `high`.
2. Where it does emit, carry `unresolvedCount`/`complete` **inside the rollup metadata**, so a consumer reading only that finding cannot miss it.
3. Report counts as a floor (">= N observed of M probed"), never a bare N.

## Related

- #781 (non-determinism across `force_refresh` runs, closed COMPLETED 2026-08-24) is the same underlying instability seen from another angle — two runs hours apart resolved different subsets.

MadaBurns:
## What happens

A throttled `check_lookalikes` run still publishes the aggregate threat rollup as a hard integer at `high` severity. Measured 2026-08-31 on openai.com (issue #865): `enumeration` said 90 probed / 12 resolved / 77 unresolved / `complete: false`, and the same payload carried **"6 lookalike domains showing pre-phishing staging signals"** at `high` with `mailCapableCount: 8` and the prose "8 can send mail in total". The `scan_status` finding said the list was a sample; the rollup — the finding a consumer keys on, at higher severity — stated the number as fact. Across eight providers the same day the unresolved rate ranged 1.2% → 86%, so side-by-side counts reported a rate-limiting artifact as a security difference between named companies.

This is the enumeration-coverage sibling of #832 (closed by #847): there the throttled instrument was the seed NS lookup and the contrary output was an ownership *verdict*; here the throttled instrument is the permutation fan-out and the contrary output is a *count with a confidence*. Neither #847's `unmeasured` law nor #781's `scan_status` sample notice reached the rollup finding itself.

## Root cause

`buildStagingSummaryFinding` / `buildMailCapableSummaryFinding` (`src/tools/lookalike-summary-findings.ts`) were fed the counted domains and the `enumeration` block, and did two things with the block: nested it under `metadata.enumeration` and flipped `confidence` from `deterministic` to `heuristic`. Nothing in the builders — or in the orchestrator that called them unconditionally whenever `mailCapableDomains.length > 0` — could abstain, and the title/prose always rendered a bare `N`.

Members whose RDAP age probe returned nothing (`registrationDays: null`, the #867 shape) were also counted with no indication that the "recent registration" corroborator had been unmeasured for them.

## Fix

`src/tools/lookalike-summary-findings.ts`

- New named constant `ROLLUP_ABSTAIN_UNRESOLVED_RATIO = 0.5` and predicate `isRollupCoverageDegraded(enumeration)`.
- New single decision point `buildThreatRollupFinding({ seedDomain, members, enumeration })` → `{ finding, abstained }`. Abstention is decided **before** either count-carrying builder is reached, so a throttled run can never produce an integer.
- New `buildRollupNotAssessedFinding` — `scan_status`, `info`, `notAssessedReason: 'enumeration_throttled'`, `inconclusive: true`, `errorKind: 'dns_error'` (the repo's codified "probe never reached the origin" markers from `lib/control-presence.ts`). Carries **no** count, no domain list; carries the enumeration stats flat plus `unresolvedRatio` and `abstainThreshold`.
- Both rollup builders now take `ThreatRollupMember[]` and, whenever they emit:
  - carry `permutationsProbed`, `candidatesResolved`, `unresolvedCount`, `complete` **flat** in their own metadata (the nested `enumeration` block stays);
  - when `complete: false`, lead the title and prose with **"At least N"**, add `countIsFloor: true`, and state the floor with its coverage numbers ("N of M probed lookups did not resolve … the true number can only be higher"); the wider "can send mail in total" figure gets the same "at least" treatment;
  - carry `ageUnknownCount` (members with `registrationDays === null`) and say "K of N could not be age-checked". Those members are still counted — the mail host is a real observation — the prose just stops implying the recency corroborator was checked.
- A complete run with all ages known renders byte-identically to before.

`src/tools/check-lookalikes.ts` (call-site only)

- The three separate collectors (`highCount`/`highDomains`/`highVerdicts`/`mailCapableDomains`) collapse into one `rollupMembers: ThreatRollupMember[]` push per threat-observed candidate; the two `if/else` rollup calls become one `buildThreatRollupFinding` call.
- An abstained run is marked `partial: true` — the same law #847 applied to the degraded ownership comparison: the throttling is transient, and `check_lookalikes` is cached for an hour, so a cached "not assessed" would outlive the condition that caused it.

## Threshold rationale

`ROLLUP_ABSTAIN_UNRESOLVED_RATIO = 0.5`, applied to `unresolvedCount / permutationsProbed`, abstain at `>=`.

- A rollup is a claim about a **set**. Once more of the probed namespace is unobserved than observed, the count is bounded below by what resolved and above by nothing: its plausible range exceeds its value, so it measures coverage, not the estate. Below the line the number is still informative and is published — as a floor, with the coverage beside it.
- Against the issue's measured table: amazon.com (1.2%), anthropic.com, perplexity.ai (39%) and cohere.com (43%) emit; microsoft.com (51%), meta.com (62%), google.com (67%) and openai.com (86%) abstain. The openai run — 12 of 90 resolved, published as "8 can send mail in total" at `high` — is the shape the constant exists to stop.
- #847 defined "degraded" as a **binary** on a different instrument (the seed's own NS lookup failed → every ownership verdict `unmeasured`). It has no ratio to align with, so this is the first threshold on the enumeration instrument rather than a second notion of the same one; the two compose (a seed-NS failure already withholds every threat observation, so no rollup member exists at all).
- Any incompleteness (`complete: false`, i.e. one unresolved lookup) triggers floor wording — that needs no threshold, it is simply what "sample" means.

## Tests

New `test/check-lookalikes-throttled-rollup.spec.ts` (8 tests, DNS mocked via `dns-mock`, tool dynamically imported per test):

- openai-shaped run (every other permutation's NS lookup rejects, ~99% unresolved) → both rollups withheld, `notAssessedReason: 'enumeration_throttled'` at `info`/`scan_status` with `inconclusive`/`errorKind`, no `lookalikeDomainCount`/`mailCapableCount`/domain lists, stats flat on the notice, no aggregate title with a leading integer;
- abstained run → `partial: true`;
- mail-capable-only (medium) rollup abstains under the same threshold;
- partially incomplete run (3 rejected lookups, well below the threshold) → "At least 1 lookalike domain showing pre-phishing staging signals", `countIsFloor: true`, `unresolvedCount: 3`, `permutationsProbed`/`candidatesResolved`/`complete: false` flat, `confidence: 'heuristic'`, no abstention notice;
- `ageUnknownCount: 1` with "1 of 1 … could not be age-checked" prose, member still counted;
- complete run → title exactly `1 lookalike domain showing pre-phishing staging signals`, no floor marker, `complete: true`, `deterministic`, not partial;
- medium rollup carries the same floor + stats contract;
- threshold unit test: the constant is 0.5, each row of the issue's table lands on the intended side, boundary at exactly 50%, and `permutationsProbed: 0` is never "degraded".

Commands run, all green:

```
npx vitest run test/check-lookalikes-throttled-rollup.spec.ts test/check-lookalikes.spec.ts \
  test/check-lookalikes-degraded-attribution.spec.ts test/check-lookalikes-recon-enrichment.integration.test.ts \
  test/lookalike-analysis.spec.ts test/lookalike-severity.spec.ts      # 6 files, 121 tests (113 baseline + 8 new)
npm run typecheck   # rc 0
npm run lint        # rc 0
```

Existing title matchers checked and unchanged: `test/check-lookalikes.spec.ts` matches `/pre-phishing staging signals/i` and `/lookalike domains? showing pre-phishing staging signals/i` — both still match the "At least N …" form, and the complete-run fixtures they use render the original title verbatim. No other source file matches the rollup title; `discover-brand-domains.ts` reads only per-domain `lookalikeDomain` metadata, `scan/post-processing.ts`'s `hasActiveImpersonation` keys on severity within a `scan_domain` run, which never includes `check_lookalikes`.

## Residuals

- `check_lookalikes` is a standalone intelligence tool outside the scored `CheckCategory` union and not in `scan_domain`'s dispatch, so **no `scan_domain` score moves**. Within the tool's own `CheckResult`, an abstained run drops the rollup's `high` penalty from its self-score — the per-domain `high` threat observations still apply.
- Per-domain findings are unchanged; an abstained run still lists each resolved candidate individually, with the existing #781 sample notice.
- Why `.com` registration ages come back `null` (#867) is being fixed separately in `lookalike-enrichment.ts`; this PR only makes the rollup honest about it via `ageUnknownCount`.
- The threshold is a first calibration from eight data points; if a future corpus shows counts stable well above 50% unresolved (or unstable below it), move the constant — everything keys on it.

Closes #865

### DIFF
--- a/src/tools/check-lookalikes.ts
+++ b/src/tools/check-lookalikes.ts
@@ -35,7 +35,7 @@ import type { CheckResult, Finding } from '../lib/scoring';
 import { buildCheckResult } from '../lib/scoring';
 import { generateCognitiveLookalikes, generateCombosquats, generateLookalikes } from './lookalike-analysis';
 import { calibrateLookalikeSeverity, type LookalikeSignals } from './lookalike-severity';
-import { classifyOwnership, type OwnershipAssessment, type OwnershipVerdict } from '../lib/ownership-attribution';
+import { classifyOwnership, type OwnershipAssessment } from '../lib/ownership-attribution';
 import { isSharedNsHost } from '../tenants/discovery/shared-ns-hosts';
 import { extractBrandName } from '../lib/public-suffix';
 import {
@@ -63,15 +63,15 @@ import {
 } from './lookalike-findings';
 import {
 	buildIncompleteEnumerationFinding,
-	buildMailCapableSummaryFinding,
 	buildNoActiveInfrastructureFinding,
 	buildNoPermutationsFinding,
 	buildNoRegisteredCandidatesFinding,
 	buildOwnershipUnmeasuredFinding,
 	buildReconCandidateFinding,
 	buildReconScanStatusFinding,
-	buildStagingSummaryFinding,
+	buildThreatRollupFinding,
 	buildTimeoutFinding,
+	type ThreatRollupMember,
 } from './lookalike-summary-findings';
 
 // ---------------------------------------------------------------------------
@@ -359,31 +359,27 @@ async function checkLookalikesCore(
 	// #264-calibrated OBSERVED-threat severity rides a separate finding that
 	// asserts nothing about ownership.
 	//
-	// `highCount` is reachable again as of Task 7b: it counts threat-OBSERVATION
-	// highs, which the ownership cap no longer touches. Under Task 7 it could
-	// never increment and the summary finding below was dead code.
-	let highCount = 0;
-	/** The counted candidates, so the aggregate summary can name what it counted (invariant 4). */
-	const highDomains: string[] = [];
-	/** Their ownership verdicts — always non-owned, since owned candidates never reach the counter. */
-	const highVerdicts = new Set<OwnershipVerdict>();
-	/**
-	 * EVERY non-owned candidate with a working mail host — a strictly WIDER set
-	 * than `highDomains`, which additionally requires a #264 corroborator.
-	 *
-	 * Tracked separately because the summary finding used to be TITLED for this
-	 * set ("N lookalike domains with mail capability detected") while being
-	 * COUNTED from the narrower one (#779). Measured on openclaw.ai the title
-	 * said 2 while six candidates in the SAME response carried `hasMX: true`;
-	 * on openclaw.org it said 1 against 3. A three-fold undercount, in the one
-	 * finding a consumer is most likely to read without expanding the per-domain
-	 * detail — and it propagated into a client-facing report.
-	 *
-	 * A domain with working mail AND a live website is not the safer case: the
-	 * site lends the mail credibility. Excluding it from a count labelled "mail
-	 * capability" was the wrong direction to be wrong in.
-	 */
-	const mailCapableDomains: string[] = [];
+	// The staging count is reachable again as of Task 7b: it counts
+	// threat-OBSERVATION highs, which the ownership cap no longer touches. Under
+	// Task 7 it could never increment and the summary finding below was dead code.
+	//
+	// EVERY non-owned, measured candidate that received a threat observation is
+	// a rollup member; `buildThreatRollupFinding` derives BOTH counted sets from
+	// the one list — the mail-capable set, and its HIGH subset. They used to be
+	// collected separately, and the summary finding was TITLED for the wider set
+	// ("N lookalike domains with mail capability detected") while COUNTED from
+	// the narrower one (#779). Measured on openclaw.ai the title said 2 while six
+	// candidates in the SAME response carried `hasMX: true`; on openclaw.org it
+	// said 1 against 3. A three-fold undercount, in the one finding a consumer
+	// is most likely to read without expanding the per-domain detail — and it
+	// propagated into a client-facing report. A domain with working mail AND a
+	// live website is not the safer case: the site lends the mail credibility.
+	//
+	// Each member carries its own `registrationDays` so the rollup can say how
+	// many of the domains it counts were never age-checked (#865) — counted
+	// still, because the mail host is a real observation; hedged, because the
+	// "recent registration" corroborator was unmeasured for them.
+	const rollupMembers: ThreatRollupMember[] = [];
 	for (const result of results) {
 		const ownership: OwnershipAssessment = ownershipByDomain.get(result.domain) ?? {
 			verdict: 'unattributed',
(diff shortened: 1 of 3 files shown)


MadaBurns:
**Full `npm test` run (this branch, worktree, Node 22):** 696 files — 690 passed, 4 skipped, 2 "failed"; 8415 tests — **8401 passed**, 13 skipped, **1 failed**. Duration 239s (import 1325s).

The two non-green files are both environmental, neither touches this change:

1. `test/subdomain-takeover-cachekey.spec.ts` — its **first** test (the one that pays the cold `import('../src/handlers/tools')`) hit the 15 s per-test timeout at 15013 ms; the two sibling tests that reuse the cached import passed. Re-run in isolation: same timeout. Re-run with `--testTimeout=60000`: **3/3 pass**. Cold-import cost on this machine, not a logic failure; `main` CI is green at the base SHA (`905a782f7`).
2. The second entry is a suite-load failure with 0 test failures — the `bv-wasm-core` module that `npm run build:wasm` produces and a fresh worktree does not have (documented in the `bv-mcp-testing` skill as expected in a fresh worktree).

Every test file that touches the lookalike surface (16 files, 328 tests) passes; `npm run typecheck` and `npm run lint` are clean.

PR #903, opened 2026-09-04
fix(lookalikes): bound the DNS-phase fan-out to the Workers connection cap so throttling stops being self-inflicted (#865 follow-up)

MadaBurns:
Follow-up to #865 / #867 / #894 / #892 (all closed — nothing here auto-closes).

## What happens

`check_lookalikes` reports most of its permutation space as unresolved on large seeds, and #892's rollup then abstains at ≥ 50 %. Measured live on prod 3.74.0 today (`force_refresh: true`):

| seed | probed | resolved | unresolved | rollup |
|---|---|---|---|---|
| openai.com | 90 | 13 | **76 (84 %)** | abstained (`enumeration_throttled`) — 4 threat observations at HIGH citing "no reachable web content" |
| cohere.com | 94 | 28 | **40 (43 %)** | published; 8 × `hasWebContent: false` |

#894 fixed the same mechanism in the enrichment phase and left this residual verbatim: *"The run's unresolvedCount … on the DNS phases is almost certainly the same mechanism."* This PR proves it, with a twist, and fixes it.

## Evidence

**Hypothesis: the DNS phases exceed the Workers six-connection cap with timers armed before dispatch. Verdict: PROVEN — with one refinement.**

1. **Code.** `filterByNsExistence` dispatched every permutation's NS query at once (`Promise.allSettled(domains.map(...))`, ~90 fetches, `PHASE1_DNS_OPTS` 2 s timer armed at call time, no semaphore). `probeWithAdaptiveBatching` dispatched 20 A/MX queries per batch (plus secondary confirmations). `dns-transport.ts` creates `AbortSignal.timeout()` *before* `guardedFetch`, so even a semaphore would not have helped — the timer runs in the queue.
2. **Resolver profile from this host (sequential, no concurrency at all).** openai.com: **11 of 90 permutations hang to the 2 s timer even queried alone** (lame/slow delegations: `oenai.com`, `oopenai.com`, `opebai.com`, …); the rest answer at p50 224 ms. cohere.com: 0 hangs. A burst of all 90 from Node: **zero HTTP 429**, the same 11 hangs. The resolver never refused the burst.
3. **Replay of the measured per-name latencies through a six-slot FIFO.** Pre-armed timers (current code): openai.com **70 unresolved** (prod 76, #865 77); cohere.com 23 (prod 40). Timers armed at dispatch: openai.com **11** (only the inherent hangs), cohere.com **0**. The refinement: it is not queue length alone — each hanging name parks one of the six slots for the *whole* 2 s window, so a dozen of them kill every slot and the queue behind them aborts unsent. That is why amazon.com (80 of 82 permutations registered, answering fast) sat at 1.2 % in #865's table while openai.com sat at 86 %.
4. **Same code on an uncapped local workerd against the same resolver:** openai.com 54 resolved / 12 unresolved (prod: 13 / 76). Local workerd does not enforce the cap; the 64-candidate gap is the platform queue.
5. **Emulation spec** (`test/check-lookalikes-dns-starvation.spec.ts`, the six-slot abortable FIFO from #894 with a per-name hold): the pre-fix fan-out shape against 90 names of which 7 hang → `peakWaiting > 0`, ≥ 15 fetches queued-then-aborted, ≥ 22 rejected. The bounded path on the same runtime → `peakWaiting === 0`, unresolved exactly 7, all `timeout`. **Mutation-checked:** `LOOKALIKE_DNS_PROBE_CONCURRENCY` 6 → 7 fails three assertions deterministically (`expected 1 to be +0`, `expected 7 to be less than or equal to 6`).
6. **Live, after** (local build via `wrangler dev`, same MCP handshake, `force_refresh: true`):

| seed | before (prod) | after, run 1 | after, run 2 |
|---|---|---|---|
| openai.com | 13 / 76 unresolved, rollup abstained | **54 / 23** (`timeout` 11, `deadline` 12), rollup published | **54 / 7** (`timeout` 7), rollup published |
| cohere.com | 28 / 40 | **39 / 0, `complete: true`** | — |
| openai `hasWebContent: false` | 4 | 2 (measured refusals) | 2 |

Caveat, stated plainly: local workerd does not enforce the six-connection cap, so the local "after" proves the pooled pipeline, the classifier and the deadlines against the real resolver; the cap behaviour is proven by the emulation spec. Prod verification after deploy is the acceptance step that remains (Residuals).

A second measurement in real workerd of the pooled phase 1 alone on the 90 openai.com names: wall **4.99 s**, 79 ok (p50 36 ms), 11 timeouts at exactly 2001 ms — the same 11 names as the host profile.

## Root cause

Unbounded DoH fan-out in `src/tools/lookalike-dns.ts` against a platform that serialises past six connections, with every timer armed at enqueue. Registered-but-lame permutations (NS answered from the parent zone, own servers dead) hold a slot for the full timer, which converts a 13 % inherent unresolved rate into 84 %. The adaptive-batching backoff could never fire (it counted rejected per-candidate promises, and the per-candidate probe settled both legs internally), and it carried a latent off-by-batch bug (advancing by the *new* size after halving, re-probing half a batch) — both now live and fixed.

Two silent-drop sites found on the way and closed: (a) the `registeredPerms.length === 0` early return reported **"No active lookalike domains detected", `deterministic`**, even when every lookup was unresolved — the enumeration stats never reached the response on that path; (b) `probeHasWebContent` mapped a timeout to `hasWebContent: false`, the HIGH corroborator (#894 residual 2).

## Fix

`src/tools/lookalike-dns.ts`
- `LOOKALIKE_DNS_PROBE_CONCURRENCY = 6` — one `mapConcurrent` pool per fan-out, counted in **queries** in flight (retries and secondary confirmation are sequential inside a query, so a pooled query never holds more than one connection). Not `SCAN_DNS_CONCURRENCY` (12): the tool is `scanIncluded: false` and the DNS phases run alone, so each owns the whole cap. Phase 2 flattens A/MX legs into the pool (a per-candidate pool of three would idle a slot whenever the legs finish apart).
- Timers armed at dispatch: the transport's per-query timer is created inside the pool worker; a per-phase caller-abort signal (`AbortSignal.timeout(remaining)`) is created at dispatch too and composed by the transport, so a deadline bounds the whole query including retries.
- `DnsPhaseOptions.deadlineMs` on `filterByNsExistence` / `probeWithAdaptiveBatching`. Victims are **counted with a reason** — `DnsProbeFailureReason = 'timeout' | 'deadline' | 'failed'` — never dropped, never "no NS". `filterByNsExistence` returns `unresolvedByReason`; a degraded `LookalikeResult` carries `probeDegradedReason`.
- `PHASE2_DNS_OPTS = { retries: 0 }` for candidate A/MX legs (default timer and secondary confirmation kept, so `hasA`/`hasMX` semantics are unchanged). A lame candidate hangs both legs; with the default retry each leg held a slot for 3 s + 3 s, and eleven such candidates spent the whole phase budget on names that would never answer. #853's case for retries is the seed's lookups, not disposable permutations.
- `detectWildcardParents` pooled and on the lean `PHASE1_DNS_OPTS` (it precedes phase 1 and used to cost up to ~7 s per lame parent zone; a failed canary only means "not a wildcard").
- `PHASE1_DNS_OPTS` and the four batching constants are byte-identical (pinned by `test/check-lookalikes.spec.ts`).

`src/tools/check-lookalikes.ts`
- Budget: `DNS_PHASES_BUDGET_MS = LOOKALIKE_TIMEOUT_MS − ENRICHMENT_RESERVE_MS − ENRICHMENT_MIN_WINDOW_MS (4 s)` = 12 s absolute from `startedAt`, so #894's enrichment window and reserve are untouchable. `NS_PHASE_BUDGET_MS = 6 s` counted from the moment phase 1 *starts* (min'd with the absolute boundary): anchoring it on `startedAt` let the seed's resilient lookups (#853, 5 s × 3) hand phase 1 an already-spent budget — measured locally as 6 resolved / 83 `deadline` one run and 54 / 27 the next. Phase 2 gets the remainder.
- `unresolvedByReason` is attached **beside** `enumeration` on the two scan_status findings (incomplete-enumeration notice; throttled rollup abstention). The five-key `enumeration` object #892 reads is byte-compatible (asserted). The builders live in `lookalike-summary-findings.ts` (not touched per the ownership brief), so the annotation is applied in the orchestrator.
- Early return with no registered candidate and unresolved lookups: emits the incomplete-enumeration notice (with `candidatesResolved: 0`), demotes "no registered candidates" to `heuristic` when some lookups did measure an absence and omits it when none did, and marks the result `partial` so the non-answer is not cached.

`src/tools/lookalike-enrichment.ts`
- `probeHasWebContent`: timeout → `true` (unknown); only a measured transport refusal → `false`. Module doc now matches behaviour.

`scan_domain`: `check_lookalikes` is a standalone intelligence tool outside the scored `CheckCategory` union and unreferenced by `scan-domain.ts`; no scan score can move and `SCORING_MODEL_VERSION` needs no bump. Not touched: `lookalike-summary-findings.ts`, `lib/ownership-attribution.ts`, `lib/maturity-staging.ts`, `packages/dns-checks`, `scan-domain.ts`, `scoring-version.ts`, `CHANGELOG.md`, version fields.

## Tests

New `test/check-lookalikes-dns-starvation.spec.ts` (13) and `test/lookalike-enrichment-web-content.spec.ts` (5), written red first (11 failed on the old code, the end-to-end guard with `expected 90 to be less than or equal to 6`):
- REPRODUCTION (pre-fix fan-out shape through the real transport) against the six-slot runtime; bounded `filterByNsExistence` on the same runtime (`peakWaiting === 0`, unresolved = exactly the hangs, all `timeout`) — the mutation guard; deadline: unissued tail counted `deadline`, `registered + unresolved === domains`; `deadline` vs `timeout` vs `failed` classification; no-deadline behaviour unchanged.
- `probeWithAdaptiveBatching`: ≤ 6 in flight for a 25-candidate set; deadline-spent → every candidate `probeDegraded` + `deadline`, no fetch issued, no synthesised `hasA`/`hasMX`; per-leg reasons; `PHASE2_DNS_OPTS` pinned with a single MX attempt for a hanging candidate.
- End to end through `checkLookalikes`: peak DoH in flight ≤ 6 on a 90-permutation seed; `enumeration` has exactly the five keys; `unresolvedByReason` sums to `unresolvedCount`; all-unresolved run → incomplete notice, `partial`, no "no active lookalikes", nothing `deterministic`; partially-unresolved run → "no registered" demoted to `heuristic` beside the notice.
- Web content: timeout → `true`; measured refusal → `false`; 200/3xx/5xx → `true`; deadline spent → `true` with no fetch; through `enrichLookalikes`, a hung HEAD cannot produce the corroborator.

Two existing assertions corrected because they pinned the defects: `test/check-lookalikes-registration-age.spec.ts` (the old fan-out shape can no longer manufacture `hasWebContent: false`) and `test/check-lookalikes.spec.ts › should handle all probe failures without crashing` (all-rejected lookups now yield the incomplete notice + `partial`, not "No active lookalike").

```
npm run typecheck                                       → 0 errors
npm run lint                                            → 0 errors (1 pre-existing warning)
npx vitest run test/check-lookalikes test/lookalike     → 11 files, 185 passed
node scripts/ci/typecheck-tests.mjs                     → OK, no test file exceeds its baseline
npm test                                                → see below
```

`npm test` (full suite, Workers pool): **8490 passed / 13 skipped / 14 failed**, 693 files passed, 6 files failed — none real:

- `test/wasm-integration.test.ts` — suite-load error `No such module …/crates/bv-wasm-core/pkg/bv_wasm_core_bg.wasm`; the documented fresh-worktree shape (`npm run build:wasm` is not run by `worktree-setup.sh`), 0 test failures inside it.
- 13 tests in `test/freemium-limits.spec.ts`, `test/index.spec.ts`, `test/oauth/{authorize-post,e2e,token}.spec.ts` — all auth/OAuth/API-key cases, none near the lookalike path. Cause: the throwaway `.dev.vars` (`BV_API_KEY`, `OWNER_ALLOW_IPS`) I had created for the local `wrangler dev` evidence run is read by vitest-pool-workers as bindings, so "OWNER_ALLOW_IPS unset" / "valid key" cases saw my values. Re-run in isolation after deleting it: **5 files, 127/127 passed**.

## Residuals

- **Prod verification.** Local workerd does not enforce the cap; the acceptance evidence for the cap itself is the emulation spec plus the prod re-run after deploy (`check_lookalikes openai.com force_refresh` — expect unresolved ≈ 11–25 of 90, `unresolvedByReason.timeout` ≈ 11, rollup published).
- **Inherent hangs.** ~11 openai.com permutations are lame delegations that hang the resolver for the full timer even when queried alone (13 %); they are now reported honestly as `timeout`, and nothing client-side can resolve them.
- **Large seeds still exhaust the 12 s DNS budget in phase 2** (0–12 `deadline` cuts per run for openai.com's 54 registered candidates) and enrichment's window (`not_attempted` tails). `LOOKALIKE_TIMEOUT_MS = 20 s` is the constraint; raising it, or leaning phase 2 further, is an operator call.
- `dns-transport.ts` arms `AbortSignal.timeout()` *before* `sem.run()`, so `scan_domain`'s semaphore path (`SCAN_DNS_CONCURRENCY = 12`, also > 6) still arms timers at enqueue — same defect class, out of scope here, worth its own issue. Also: its `AbortError` branch never matches `AbortSignal.timeout`'s `TimeoutError`, so real timeouts surface as `DNS query failed: The operation was aborted due to timeout` (retry still happens via the generic branch); the classifier here accepts both spellings.
- `check_lookalikes` is still declared `tier: 'protective'` + `scanIncluded: false` (pre-existing, noted in #894).

### DIFF
--- a/src/lib/dns-transport.ts
+++ b/src/lib/dns-transport.ts
@@ -49,11 +49,14 @@ function retryDelay(attempt: number): Promise<void> {
  *
  * @param token - Optional auth token sent as `X-BV-Token` (for custom secondary resolvers).
  * @param useEdgeCache - If true, attaches Cloudflare `cf` cache directive. Omit for external origins.
+ * @param signal - Optional caller abort signal, composed with the per-fetch
+ *   timeout exactly as `queryDns` composes it for the primary fetch, so a
+ *   caller deadline bounds this fetch too. Absent → timeout alone (unchanged).
  */
 export async function fetchDohOutcome(
 	url: string,
 	timeoutMs: number,
-	opts?: { token?: string; useEdgeCache?: boolean; semaphore?: Semaphore },
+	opts?: { token?: string; useEdgeCache?: boolean; semaphore?: Semaphore; signal?: AbortSignal },
 ): Promise<DohOutcome> {
 	try {
 		const headers: Record<string, string> = { Accept: 'application/dns-json' };
@@ -62,11 +65,11 @@ export async function fetchDohOutcome(
 			fetch(url, {
 				method: 'GET',
 				headers,
-				signal: AbortSignal.timeout(timeoutMs),
+				signal: opts?.signal ? AbortSignal.any([AbortSignal.timeout(timeoutMs), opts.signal]) : AbortSignal.timeout(timeoutMs),
 				redirect: 'manual',
 				...(opts?.useEdgeCache ? { cf: { cacheTtl: DOH_EDGE_CACHE_TTL, cacheEverything: true } } : {}),
 			});
-		const response = opts?.semaphore ? await opts.semaphore.run(doFetch) : await doFetch();
+		const response = opts?.semaphore ? await opts.semaphore.run(doFetch, opts.signal) : await doFetch();
 		if (!response.ok) {
 			const status = response.status;
 			await disposeUnreadResponseBody(response);
@@ -170,9 +173,7 @@ async function queryDnsUncached(domain: string, type: RecordTypeName, dnssecChec
 		// Either firing aborts the fetch. AbortSignal.any is a standard Web API
 		// available in workerd; falls back gracefully if the caller didn't pass
 		// a signal (timeout alone).
-		const fetchSignal = callerSignal
-			? AbortSignal.any([AbortSignal.timeout(timeoutMs), callerSignal])
-			: AbortSignal.timeout(timeoutMs);
+		const fetchSignal = callerSignal ? AbortSignal.any([AbortSignal.timeout(timeoutMs), callerSignal]) : AbortSignal.timeout(timeoutMs);
 
 		try {
 			response = await guardedFetch(url, {
@@ -226,10 +227,20 @@ async function queryDnsUncached(domain: string, type: RecordTypeName, dnssecChec
 		const data = validated.data as DohResponse;
 
 		if (confirmWithSecondaryOnEmpty && !opts?.skipSecondaryConfirmation && !hasTypedAnswers(data, type)) {
-			const secondaryOpts = opts?.secondaryDoh
-				? { secondaryDoh: { url: opts.secondaryDoh.endpoint, token: opts.secondaryDoh.token } }
-				: undefined;
+			// The caller's signal rides into the secondary fetches too (PR #903
+			// review): without it, an empty primary answer let the confirmation
+			// arm its own fresh timer and run up to `timeoutMs` PAST the caller's
+			// deadline while holding a connection slot.
+			const secondaryOpts = {
+				...(opts?.secondaryDoh ? { secondaryDoh: { url: opts.secondaryDoh.endpoint, token: opts.secondaryDoh.token } } : {}),
+				...(callerSignal ? { signal: callerSignal } : {}),
+			};
 			const secondaryResult = await confirmWithSecondaryResolvers(domain, type, dnssecCheck, timeoutMs, sem, secondaryOpts);
+			// An aborted confirmation is NOT a confirmed-empty answer: the caller
+			// stopped the measurement, so it must see the abort, never `data`.
+			if (callerSignal?.aborted) {
+				throw new DnsQueryError(`DNS query aborted by caller`, domain, type);
+			}
 			if ('kind' in secondaryResult && secondaryResult.kind === 'unconfirmed') {
 				// Secondary confirmation unavailable — keep the primary result as-is.
 				// (Do NOT change primary to empty; primary is authoritative when we can't verify.)
@@ -259,18 +270,19 @@ export async function confirmWithSecondaryResolvers(
 	dnssecCheck: boolean,
 	timeoutMs: number,
 	sem?: Semaphore,
-	opts?: { secondaryDoh?: { url: string; token?: string } },
+	opts?: { secondaryDoh?: { url: string; token?: string }; signal?: AbortSignal },
 ): Promise<DohResponse | { kind: 'unconfirmed' }> {
 	const bvDnsUrl = opts?.secondaryDoh ? buildDohUrl(opts.secondaryDoh.url, domain, type, dnssecCheck) : null;
 	const googleUrl = buildDohUrl(GOOGLE_DOH_ENDPOINT, domain, type, dnssecCheck);
+	const signal = opts?.signal;
 	const candidates = [
 		bvDnsUrl
-			? fetchDohOutcome(bvDnsUrl, timeoutMs, { token: opts!.secondaryDoh!.token, semaphore: sem })
+			? fetchDohOutcome(bvDnsUrl, timeoutMs, { token: opts!.secondaryDoh!.token, semaphore: sem, signal })
 			: Promise.resolve({ kind: 'error', reason: 'network' } as const),
 		// No edge cache on the secondary: it's the rare fallback used precisely when
 		// the primary missed, so caching its responses risks persisting a transient
 		// empty at the edge for the cache TTL (#199).
-		fetchDohOutcome(googleUrl, timeoutMs, { semaphore: sem }),
+		fetchDohOutcome(googleUrl, timeoutMs, { semaphore: sem, signal }),
 	];
 	const results = await Promise.allSettled(candidates);
 	for (const r of results) {
--- a/src/tools/check-lookalikes.ts
+++ b/src/tools/check-lookalikes.ts
@@ -48,6 +48,7 @@ import {
 	queryPrimaryMx,
 	queryPrimaryNs,
 	type LookalikeResult,
+	type UnresolvedByReason,
 } from './lookalike-dns';
 import { EMPTY_RDAP_PROBE, enrichLookalikes, probePrimaryRegistration } from './lookalike-enrichment';
 import {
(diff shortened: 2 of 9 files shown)


MadaBurns:
Review fix pushed as `4c694da6d`: the caller signal now rides into `confirmWithSecondaryResolvers` / `fetchDohOutcome` (composed with the per-fetch timeout exactly like the primary; handed to the semaphore too), and `queryDnsUncached` throws `aborted by caller` after an aborted confirmation so it can never surface as a confirmed-empty answer. Absent signal → behaviour unchanged. Tests: `test/dns-transport-signal.test.ts` (+2) and the starvation spec (+1: empty-answer candidate with a hanging secondary is cut at `deadlineMs`, counted `deadline`, phase returns within budget). Mutation check: dropping the signal from the Google secondary fetch turns both red at ~3.0 s. Gates: `npx vitest run test/check-lookalikes* test/lookalike* test/dns-transport* test/dns*.spec.ts` → 20 files, 306 passed; typecheck 0 errors (after rebuilding `packages/dns-checks/dist` for the merged #901 symbols); lint clean.

PR #953, opened 2026-09-09
fix(deploy): block deploy:prod when a sidecar Worker is behind its source (#945)

### LINKED ISSUE #945: bv-whois deployed Worker is stale (last deploy 2026-05-20, predates #526): deploy after #935 and add a deploy:prod post-step / checklist line so it cannot drift again
## Summary

The live `bv-whois` Worker is stale, and nothing in the repo's deploy path would have caught it. `npm run deploy:prod` deploys the MCP Worker only; `packages/bv-whois` has its own `wrangler deploy` script that no root script, workflow, or checklist invokes — the same gap CLAUDE.md already documents for `bv-infra-probe`.

Measured 2026-09-09 from `packages/bv-whois`: `npx wrangler deployments list` shows the newest deployment at **2026-05-20T22:01Z** (10 deployments total, all 2026-05-14 → 2026-05-20). #526 — which added WHOIS-sourced dates and registrant privacy to the shim payload — merged **2026-07-20**. PR #935's implementer observed the same from the other side: the live bundle returns only `{registrar, registrarIanaId, source}`, no dates, so prod `.dk` lookups fall through to `whois_error`.

- `package.json:46` — `deploy:prod` = `check:deploy-freshness && check:release-integrity && build dns-checks && check:bindings && inject-private-config && brand-audit-schema-preflight && check:bindings:prod && wrangler deploy --config wrangler.production.jsonc`. No `bv-whois` step. `grep -n whois package.json` → nothing.
- `scripts/inject-private-config.cjs` — `grep -in whois` → nothing (it only merges bindings for the main Worker; `BV_WHOIS` is declared in `wrangler.jsonc:86-87` as a service binding to `bv-whois`).
- `.github/workflows/*.yml` — `grep -n bv-whois` → nothing.
- `packages/bv-whois/package.json` — `"deploy": "wrangler deploy"` exists but is operator-manual.
- Last source change to `packages/bv-whois` on main: dependency bumps (#881/#883 merges) after #526's `lookup.ts` changes — none deployed.
- CLAUDE.md → "`npm run deploy:prod` does NOT deploy bv-infra-probe — deploy it explicitly" — identical class.

## Proposed handling

(a) **One-off**: after #935 merges, `cd packages/bv-whois && npx wrangler deploy` (either deploy order is safe per #935's compatibility notes), then verify with a `.dk` `rdap_lookup` against prod that `source` is `whois`/`redacted` and dates are populated.

(b) **Stop the drift**: pick one —
- add a `deploy:whois` root script (`npm -w packages/bv-whois run deploy`) and a `deploy:prod` post-step that runs it **only when `packages/bv-whois` changed since the last deployed tag** (a freshness check like `check:deploy-freshness` scoped to that path), or
- add the line to the release checklist in the `bv-mcp-release` skill and to CLAUDE.md's Deployment section next to the bv-infra-probe caveat, and have `check:deploy-freshness` warn when `packages/bv-whois/src` has commits newer than the shim's last deployment (readable via `wrangler deployments list`).

The second is cheaper and matches how bv-infra-probe is handled today; the first prevents the same report a third time.

## Refs

- #526 (the undeployed change), #931 / #935 (where the staleness was discovered; #935 body §"The deployed bv-whois Worker is stale")
- CLAUDE.md → Deployment / Bindings ("does NOT deploy bv-infra-probe")

MadaBurns:
Closes #945.

## The defect

`npm run deploy:prod` deploys the MCP Worker only. The two sidecar Workers have their own configs and their own deploy commands, and **nothing in the repo invoked them** — so their drift was invisible in an otherwise-green deploy. The existing mitigation was a `CLAUDE.md` caveat telling the operator to remember, which is exactly what failed.

Measured 2026-09-09:

| Sidecar | Live deployment | Newer source commits |
|---|---|---|
| `bv-whois` | 2026-05-20T22:01:09Z | **4** — incl. #526, the commit that added the WHOIS dates the RDAP fallback asks it for |
| `bv-infra-probe` | 2026-05-20T21:27:44Z | **7** — incl. #803, #824, #827, #837, #838 |

`bv-infra-probe` was not in the original issue; the gate found it.

## The gate

`check:sidecar-freshness` runs early (~2s) in `deploy:prod`, `deploy:prod:staged` and `deploy-private.mjs` — the second deploy door, which skips the git freshness gate entirely. It reads each sidecar's live deployment via `wrangler deployments list --json` and compares against commits touching that sidecar's source.

**It blocks.** A warning inside an `&&` chain ending in `wrangler deploy` scrolls off the top of a long log, which is how this drifted for three months.

| Situation | Verdict | Exit |
|---|---|---|
| no newer source commit | `fresh` | 0 |
| newer commits | `drifted` — names every commit + fix command | 1 |
| wrangler unauth / offline / missing | `unverified` | 1 |
| stdout is banner text, not JSON | `unverified` | 1 |
| stdout is `[]` | `unverified` — **explicitly never "fresh"** | 1 |
| `git log` fails | `unverified` | 1 |
| `BV_ALLOW_STALE_SIDECARS=1` | `override` — prints the drift it bypasses | 0 |

Fail-closed by construction: a bad token puts a 199-byte banner on **stdout** alongside a non-zero status, so exit status is checked and stdout is never parsed without a guard. The override is deliberately distinct from `BV_ALLOW_STALE_DEPLOY` — bypassing the git gate for a rollback must not silently also ship a stale sidecar.

## Design notes

- Structured like `check:deploy-freshness`: a **pure decision core** with no `node:*` imports (so it is testable in the Workers pool, where importing `node:child_process` is a hard SIGSEGV) plus a thin CLI shell owning the `spawnSync` calls.
- `bv-infra-probe` watches `src/lib/authoritative-dns-infra` as well as `infra-probe.ts` — a change there ships to the MCP Worker and to the probe *never*, which is this exact drift class.
- Requiring wrangler auth narrows nothing: `deploy:prod` already cannot complete without it, because `brand-audit-schema-preflight.mjs` shells out to `wrangler --remote` and hard-fails.
- `parseNewestDeployment` takes the **max** `created_on`, not `[0]` — the API returns ascending order, so a `[0]` read would take the oldest and always report stale. Pinned by a test.
- An SSOT audit asserts `SIDECAR_TARGETS` covers every tracked `wrangler*.jsonc` that is not the main Worker, so a third sidecar cannot be silently omitted later.

## ⚠️ Operator note

The next `deploy:prod` **will block** until both sidecars are deployed (`npm run deploy:whois`, `npm run deploy:infra-probe`) or the override is set. That is the intended effect. Deploying the sidecars is operator-run and not part of this PR.

## Verification

- `test/sidecar-deploy-drift.spec.ts` (Workers pool, 23 tests) + `test/audits/sidecar-deploy-drift-check.node.test.ts` (Node pool, 10 tests) — incl. an exhaustive fail-closed sweep over {fresh, drifted} x {probe ok, probe failed} x {override on/off} asserting any `ok: true` implies `code in {fresh, override}`, and a regression fixture built from the real measured numbers.
- Full audit suite: `104 passed | 1 skipped`, `670 tests passed`.
- `npm run typecheck` RC=0; `npm run typecheck:tests` ratchet `TOTAL: 735 (baseline 735, delta +0)`.
- Gate exercised both ways: blocking run exits 1 naming all 11 drifted commits; `BV_ALLOW_STALE_SIDECARS=1` exits 0 while printing the same drift.

No changes to `src/` or `packages/`; no scoring impact.


MadaBurns:
## Adversarial review — one real finding, fixed in fb57173

The review found a **fail-open in the fail-closed gate** — the exact defect class this PR exists to eliminate. Fixed rather than waived.

### The finding

The sidecar gate's git evidence is `git log HEAD -- <watchPaths>`, which can only see commits that are **ancestors of HEAD**. The in-code justification for comparing against `HEAD` rather than `origin/main` is that `check:deploy-freshness` has already proven `HEAD ⊇ origin/main` by the time it runs.

That holds for `deploy:prod` and `deploy:prod:staged`, whose npm chains run freshness first. **It did not hold for `scripts/deploy-private.mjs`** — that door never ran `check:deploy-freshness` at all, and this PR wired the sidecar gate into it as the very first step. The gate was resting on a precondition nothing enforced there.

**Failure scenario:** an operator runs `deploy-private.mjs` from a checkout behind `origin/main`. A commit touching `packages/bv-whois/src` or `src/lib/authoritative-dns-infra` has landed upstream but is not in local HEAD → `git log HEAD` cannot see it → drift list comes back **empty** → the gate prints `Sidecar deploy-drift OK`. A false green, on the one door this PR's own audit test claims is now covered.

Stale checkouts are a proven incident class in this repo: the `deploy-freshness.ts` docstring cites 2026-07-27, where a stale-tree deploy exited green and silently reverted three merged commits.

### The fix

`deploy-private.mjs` now runs `check:deploy-freshness` before the sidecar gate, making the documented precondition **true on all three doors** rather than assumed on two and absent on the third. This also closes a pre-existing gap — that door skipped freshness entirely.

The module comment now states the dependency explicitly, including what to do when wiring a fourth caller. Three new audit assertions pin the ordering on all three doors.

**Verified by negative control:** reverting the freshness step fails *exactly* the new private-door test (`1 failed | 18 passed`); restored, `19 passed`. The test is not decoration.

### Verified clean (no other findings)

The reviewer worked from the branch's raw file contents rather than the local checkout, having correctly noticed the shared workdir had drifted to another concurrent branch. Confirmed by reading: every fail-open path (unparseable stdout, empty array, non-zero wrangler exit, missing/NaN `created_on`, empty probe set, failing `git log`) maps to `unverified` → block, with the ~199-byte auth-banner hazard pinned by a fixture using the real banner text; `parseNewestDeployment` takes `Math.max` (not `[0]`, not `[-1]`), pinned by an out-of-order test; `selectCommitsAfter` uses a strict `>` boundary; the override is checked first, prints what it bypasses, and `BV_ALLOW_STALE_DEPLOY=1` provably does **not** release it; `spawnSync` uses argv arrays with no `shell: true` and no value interpolated from wrangler or git *output*; the SSOT audit runs a real `git ls-files` and would genuinely fail on a forgotten third sidecar; and `src/workers/infra-probe.ts` does import from `src/lib/authoritative-dns-infra`, so that watch path is real and not permanent false drift.

*Caveat:* this reviewer had no shell access, so the above is verified by reading, not execution. The gate's own behaviour was separately exercised by me in both modes earlier (blocking run exits 1 naming all 11 drifted commits; `BV_ALLOW_STALE_SIDECARS=1` exits 0 while printing the same drift).

**One sub-threshold note, not actioned:** `bv-whois`'s watch path `packages/bv-whois/src` includes `__tests__/`, so a test-only commit counts as drift. Over-inclusive in the fail-closed direction — deliberate, and cheaper than maintaining an exclusion list.

Issue #900, opened 2026-09-03
Package abstentions in check-caa / check-mx / check-ns return checkStatus 'error' with score 100 — never retried, cached, and misreadable as a pass


MadaBurns:
## Summary

Four package abstention paths return `checkStatus: 'error'` **with `score: 100`**, so `scan_domain`'s transient-zero retry never fires for them and the abstention is cached as if it were a stable measurement.

Sites (`packages/dns-checks/src/checks/`, origin/main @ 85e63ee8d):

- `check-caa.ts` — the `delegationStatus === 'unknown'` early return and the `lookupCaa` catch (two sites)
- `check-mx.ts` — the `queryDNS(domain, 'MX')` catch
- `check-ns.ts` — the NS-query catch (the two other `check-ns.ts` abstentions already return `score: 0, passed: false, partial: true` and are fine)

Each builds `buildCheckResult(cat, [info "not assessed"])` — an info-only finding set scores **100** — then spreads `checkStatus: 'error'` over it, without `partial: true`.

## Why it matters

- `shouldRetry` in `src/tools/scan-domain.ts` is `checkStatus === 'error' && score === 0`. These results are excluded from scoring (correct) but **never retried**, unlike `safeCheck` throws and the mta_sts abstention shipped in #898 — a transient resolver blip on caa/mx/ns is permanently unmeasured for that scan.
- No `partial: true` → the abstention is written to the 5-minute scan cache and re-served.
- The `score: 100` next to `checkStatus: 'error'` is the shape CLAUDE.md warns against: a consumer that reads `score`/`passed` without `checkStatus` (bv-web-prod consumes these package exports directly) sees a perfect score for a control nobody measured. The repo's standard non-answer contract is `checkStatus: 'error'` + `score: 0` + `partial: true` (`buildDnsErrorResult`, `check-mta-sts.ts` post-#898, the two already-correct `check-ns.ts` sites).

## Proposed

Route the four sites through one shared `buildNotAssessedResult(category, finding, status)` helper in `check-utils.ts` that emits `score: 0, passed: false, partial: true` alongside `checkStatus`, and pin the contract with a package-level audit test over every `checkStatus: 'error' | 'timeout'` return in `checks/`. Scoring is unaffected (excluded either way), so no `SCORING_MODEL_VERSION` bump — but the dns-checks package/corpus version moves per the lockstep rule.

Surfaced while fixing #889 (PR #898 residual 1).

[...]

Issue #638, opened 2026-08-07
Our own Cloudflare WAF blocks our own scanner: blackveilsecurity.com scores 100/A+ on 17 of 19 checks


MadaBurns:
Found by a live post-deploy accuracy sweep of prod 3.43.0.

## Observed

Two checks against our own flagship domain are blocked by our own Cloudflare WAF:

```
check_http_security(blackveilsecurity.com)
  -> "Cloudflare WAF challenge intercepted", httpStatus 403, checkStatus: error, score 0

check_mta_sts(blackveilsecurity.com)
  -> "Cloudflare WAF challenge intercepted - policy accessibility inconclusive"
     https://mta-sts.blackveilsecurity.com/.well-known/mta-sts.txt, httpStatus 403
```

Consequently `scan_domain(blackveilsecurity.com)` reports:

```json
"checkStatuses": {"mta_sts": "error", "http_security": "error"},
"inconclusiveCategories": ["mta_sts", "http_security"],
"evidence": {"attempted": 19, "completed": 17, "ratio": 0.894...},
"score": 100, "grade": "A+"
```

## Why it matters

1. **The advertised 100/A+ rests on 17 of 19 checks** — and the two unmeasured ones include HTTP security headers. The scoring behaviour is *correct* (inconclusive is excluded rather than zeroed, per the evidence-gate design) and `evidence.ratio` reports the shortfall honestly. But a marketing-visible perfect score that partly depends on unmeasurability is worth an explicit decision.
2. **Any Cloudflare-protected customer hits the same wall.** This is not specific to us — it is a systematic blind spot against a large slice of the addressable market.

## Secondary defect: contradictory flags

Both WAF-blocked responses set **both**:

```json
"inconclusive": true, "missingControl": true
```

Those are mutually exclusive by the model's own semantics — `missingControl` means "deterministically absent" (and zeroes the category), `inconclusive` means "could not be measured" (and excludes it). Today only the inconclusive path is taken, so the contradiction is latent; if that ordering ever changed, a WAF challenge would zero a category as though the control were genuinely missing.

## Suggested directions

- allowlist the scanner's egress at our own WAF so we can measure our own domain
- treat a WAF challenge as explicitly inconclusive and never set `missingControl` alongside it
- surface the evidence ratio next to any published score so "100/A+" is not read as "19/19 measured"

MadaBurns:
## Status: the code defect named here is fixed; the two operator-facing items remain

[...]

PR #956, opened 2026-09-09
fix(scoring): subdomain_takeover and dkim abstain when every probe fails (#948)

### LINKED ISSUE #948: Transient probe failures scored as verdicts: subdomain_takeover clean-100 (cached), DKIM high/50, dane_https/svcb_https 95, subdomailing/txt_hygiene 100
## Summary

Sibling class to #900. #900 was "a check abstains in the wrong SHAPE"; this is "a check does not abstain at all": a thrown DNS/HTTP probe is swallowed and the check returns a COMPLETED result (no `checkStatus`) carrying a confident verdict — a clean pass or a scored deficiency — for a measurement that never happened. Scoring cannot exclude it (`checkStatus` is what the engine gates on), `scan_domain`'s transient-zero retry cannot fire, and because nothing sets `partial`, the verdict is written to the 5-minute per-check cache.

Surfaced by the runtime abstention audit in #946 (`packages/dns-checks/src/__tests__/checks/abstention-shape.audit.test.ts`, `test/audits/check-abstention-shape.audit.test.ts`), which drives every check with a resolver/fetch that always throws. The audit asserts only the implication `checkStatus ⇒ not-assessed shape`, so these pass it vacuously; they are recorded there as the "completes with a verdict" set.

## Sites (origin/main @ 2ea9df099), worst first

1. **`packages/dns-checks/src/checks/subdomain-takeover-analysis.ts:468-470`** — the outer `catch { /* No CNAME or query failed; not critical. */ }` swallows a thrown CNAME lookup. Under a throwing resolver `check_subdomain_takeover` returns `"No dangling CNAME records found"` — `score: 100, passed: true`, no `checkStatus`, no `partial` — and that clean verdict is CACHED. A resolver outage reads as "no takeover vectors". (Reviewer of #946 confirmed via a dist probe.) The Worker wrapper's `markProbeInconclusive` handles a cut HTTP fingerprint probe correctly (#946 adds `partial` there), but a thrown DNS query never reaches it.
2. **`packages/dns-checks/src/checks/check-dkim.ts:44-49`** — `probeSelectorWithCname` swallows the TXT throw (`records = []`), so every selector reads as absent and the check emits `"No DKIM records found"` at `high` (score 50, `passed: true`). Scan-visible: DKIM is a core-tier category (10 in `mail_enabled`), so a resolver blip penalises a correctly-configured domain by up to −25 on that category.
3. `check_dane_https` / `check_svcb_https` — thrown TLSA / HTTPS-RR lookups read as absent, scored `low` (95).
4. `check_subdomailing` — `extractSpfIncludeChain` swallows the throw and the check reports a clean 100 (documented as deliberate in `transient-inconclusive.test.ts`, but it is still a clean verdict from a probe that never completed).
5. `check_txt_hygiene` — clean 100 under a throwing resolver.

Non-scored tools that return a completed clean result under a throwing resolver (lower stakes, no scan impact): `check_agent_discovery`, `check_dbl`, `check_dnssec_chain` (`partial` but no `checkStatus`), `check_fast_flux`, `check_lookalikes`, `check_mx_reputation`, `check_nsec_walkability`, `check_rbl`, `check_shadow_domains`, `check_srv`, `check_zone_hygiene`.

## Why this is an operator call, not a drive-by fix

Every fix here MOVES scan scores: converting a swallowed throw into `buildNotAssessedResult` changes a category from "scored 50/95/100" to "excluded and renormalised" for every scan that hits a transient failure — the opposite direction of #900's fixes, which only touched already-excluded categories. Per the release skill's lockstep rule that is a `SCORING_MODEL_VERSION` bump with a history entry plus a package/corpus bump, and DKIM in particular re-grades customers. Recommend taking (1) and (2) first: (1) is a false clean on the one category whose whole point is catching an absence, and (2) is a false penalty on a core category.

The contract the fixes should land in is already pinned: `buildNotAssessedResult` (package) / `buildDnsErrorResult` (Worker), and the #946 audits will assert the shape the moment a site starts emitting `checkStatus`.

Refs #900, #946

MadaBurns:
Partially closes #948 — the two scan-visible sites. See **Scope** below.

**Scoring change — `SCORING_MODEL_VERSION` 1.25.0 → 1.26.0.** `@blackveil/dns-checks` 1.37.0 → 1.38.0 (`PARITY_CORPUS_VERSION` in lockstep) — bv-web-prod re-vendor required.

## The defect

Sibling class to #900. #900 was "a check abstains in the wrong **shape**"; this is "a check does not abstain **at all**". Both checks swallowed a thrown DNS query and returned a COMPLETED result — no `checkStatus`, no `partial` — carrying a confident verdict for a measurement that never happened. So the engine could not exclude the category (`checkStatus` is what it gates on), `scan_domain`'s transient-zero retry could not fire, and the false verdict was written to the 5-minute per-check cache.

| Check | Under a throwing resolver, before | After |
|---|---|---|
| `subdomain_takeover` | `"No dangling CNAME records found"`, score **100**, `passed: true`, **cached** | n/a (excluded) |
| `dkim` | `"No DKIM records found"` at `high`, score **50** | n/a (excluded) |

`subdomain_takeover` handed out a clean bill of health from the one check whose entire purpose is catching an absence. `dkim` is **core-tier** (weight 10 in `mail_enabled`), so a resolver blip cost a correctly-configured domain up to **−25** on that category.

## The fix

Both return `buildNotAssessedResult(..., 'error')` — score 0, `passed: false`, `partial: true`, `checkStatus: 'error'` — so the category is renormalised out of the weighted score and shown n/a rather than penalised or passed.

`'error'` and **not** `'timeout'` is load-bearing: `shouldRetry()` is error-only, so a `'timeout'` abstention would never self-heal within a scan. The findings are `info`, carry `inconclusive` + `errorKind: 'dns_error'`, and deliberately never `missingControl` (#638 law) — both #946 audits now pass **non-vacuously** for these checks for the first time.

## The rule is ZERO-answered, not a ratio

A **partial** failure still emits the existing verdict at its existing severity and score, and merely narrows its metadata to what actually answered — `selectorsChecked` becomes the answered selectors plus a new `selectorsUnmeasured`, and the clean takeover verdict gains `subdomainsUnmeasured`. A DKIM key found in **any** selector still returns a completed result no matter how many siblings failed: positive evidence is monotone and an unmeasured sibling cannot invalidate it.

A ratio threshold (e.g. "abstain below 60% answered") was considered and **rejected**: there is no measured per-query DoH throw rate for this fleet, so the number would have been picked from intuition while silently setting how often a core-tier category abstains in production. That is the unmeasured-threshold trap, and it can be added later with data.

## Cross-repo seam

`scanSubdomainForTakeover` keeps its exported signature **byte-identical** — bv-web-prod vendors this package as a tarball — so the new failure signal rides on an additive internal function. The 8 existing call sites in its spec needed no change (verified against the rebuilt `.d.ts`).

## Direction note

This **reverses direction** relative to 1.24.0/#900, which only reshaped already-excluded categories. This one newly **excludes** categories that were previously scored. Recorded plainly in the `SCORING_MODEL_VERSION` history entry. No domain whose probes answer changes by a single point.

Population cannot be sized from the corpus — the defect made affected scans indistinguishable from clean ones, which is precisely the problem.

## Scope

Two scan-visible sites only. `dane_https`, `svcb_https`, `subdomailing` and `txt_hygiene` carry the same defect at lower stakes and **remain open on #948**, so the DKIM re-grade is attributable on its own rather than mixed into five other movements.

## Tests

- `test/check-subdomain-takeover.spec.ts:177` pinned the bug as correct (1 info finding, score 100) — rewritten to assert the abstention shape.
- Added: only-some-probes-fail keeps the clean verdict with `subdomainsUnmeasured`; four DKIM boundary cases (all-throw → abstain; explicit-selector throw → abstain; one throws → `high` retained at score 50 with narrowed metadata; key found despite failures → completed).
- Corrected a now-false comment in `test/scan-domain.spec.ts` ("DKIM is excluded because it swallows DNS errors internally").
- `Tests 172 passed` + `159 passed` + `51 passed`; `npm run typecheck` clean; parity corpus contract green with only the version bump.


MadaBurns:
## Adversarial review — no functional defects; one false claim corrected (5e1e6d5)

The reviewer worked from the live PR diff rather than the local checkout, having noticed the shared workdir had drifted to a concurrent branch — the right call, and worth noting for anyone reproducing.

### Corrected: a comment that stated something untrue about a cross-repo seam

The reviewer flagged (at ~40 confidence, below their reporting bar) that the justification on the `scanSubdomainForTakeover` wrapper asserts the name is barrel-exported, appears in the published `.d.ts`, and is compiled against by bv-web-prod through the vendored tarball. **They were right that this is false**, and I verified it:

- `scanSubdomainForTakeover` is **not** re-exported from `packages/dns-checks/src/index.ts`
- the manifest's `exports` map offers only `.`, `./scoring`, `./whois`, `./cert` — **no subpath reaches** `checks/subdomain-takeover-analysis`

So no downstream tarball consumer can import the symbol, and its signature could not break one. This is precisely the trap fleet doctrine names — *verify a vendored package's `exports`, never infer the public surface from a filename* — and the comment would have taught the next reader a false constraint about a cross-repo seam.

**The wrapper is kept**, but for the smaller honest reason: eight call sites in `test/subdomain-takeover-analysis.spec.ts` use that signature, and churning them to destructure `.findings` would add diff noise to a scoring fix for no benefit. The corrected comment says so, and says plainly that deleting the name later is safe. Comment only; 48 tests pass unchanged.

### Verified clean — including the one that would have been catastrophic

**Could a legitimate NXDOMAIN trigger the abstention?** No. Traced through `queryDnsUncached` → `queryDnsRecords` → `makeQueryDNS`: an NXDOMAIN or NOERROR-empty-answer returns `{ Answer: [] }` and maps to `[]` **without throwing**. `answered` / `cnameQueryFailed` are cleared only by a genuine transport failure (`DnsQueryError`). Independently corroborated by the existing *unmodified* test that mocks `queryDNS → []` and still gets the normal `high` "not found" verdict rather than an abstention.

That was the failure mode worth fearing: if a successful NXDOMAIN had cleared `answered`, DKIM would have abstained on **every domain that simply has no DKIM** — silently removing a core category from scoring fleet-wide.

Also confirmed by reading: every `findings.push` sits behind `if (result.records.length > 0)`, so `findings` is provably empty at the abstention point; both call sites pass `'error'` (matching `shouldRetry`'s error-only condition); neither abstention finding sets `missingControl`; `applyProviderDkimContext` only touches a `high`-severity `/No DKIM records found/i` finding, so the `info` abstention passes through with `checkStatus`/`partial` intact; `markProbeInconclusive` cannot fire on the zero-CNAME path because `fetchFn` is never reached; post-processing's downgrades only touch `critical`/`high`; and all three version surfaces moved together.

**On the rewritten test:** the reviewer pulled the actual patch and confirmed the old case was literally named `handles outer CNAME query failure gracefully` while asserting the *false clean verdict* as correct. The rewrite asserts the real contract, and the DKIM test file diff is `+61 / -0` — no assertion was removed or weakened anywhere.

*Caveat:* this reviewer had no shell access, so the above is verified by reading and tracing, not execution. The build, the parity-corpus contract, and the specs were run separately — `172 passed`, `159 passed`, `51 passed`, typecheck clean.

Issue #930, opened 2026-09-08
check_zone_hygiene: wildcard zone yields 10 medium 'internal subdomain resolves publicly' findings for hosts that do not exist (futuresoft.dk)


MadaBurns:
## Summary

`check_zone_hygiene` probes a list of sensitive hostnames (vpn, admin, staging, dev, test, corp, intranet, internal, portal, owa, ...) and reports each one that resolves as a medium "Internal subdomain resolves publicly". It does not consult wildcard detection, so on a zone with a wildcard A record every probed name "resolves" and the tool emits 10 medium findings plus an "Excessive internal subdomain exposure" finding for hosts that do not exist.

Measured 2026-09-08, engine 1.36.0 / model 1.24.0 / config `a89c66e7`.

## Reproduction

`futuresoft.dk` publishes `*.futuresoft.dk A 46.30.215.94` (DNSSEC-signed). `scan_domain` already knows this: its `ns` check emits "Wildcard DNS detected ... make subdomain takeover detection unreliable" (medium).

`check_zone_hygiene futuresoft.dk` returns `passed: false, score: 0` with:

```
Internal subdomain resolves publicly: vpn.futuresoft.dk      -> 46.30.215.94
Internal subdomain resolves publicly: admin.futuresoft.dk    -> 46.30.215.94
Internal subdomain resolves publicly: staging.futuresoft.dk  -> 46.30.215.94
... (dev, test, corp, intranet, internal, portal, owa — all 46.30.215.94)
Excessive internal subdomain exposure (10 found)
```

Control, both DoH vantages (dns.google, cloudflare-dns.com):

```
zz-cannot-exist-9f3a.futuresoft.dk  A  -> 46.30.215.94   (Status 0)
```

A name that cannot exist resolves identically, so every one of the ten findings is wildcard-synthetic. All resolve to the apex address, which is another tell.

## Why it matters

Ten medium findings and a category score of 0 on a domain that has zero such hosts. This is exactly the finding class that gets pasted into a client report as "your VPN and OWA are exposed".

## Suggested fix

- Before the sensitive-name sweep, resolve a random impossible label under the zone. If it answers, mark the zone wildcard and either skip the sweep or downgrade every hit whose address equals the wildcard answer to `info` with "wildcard-synthetic" in the detail.
- Reuse the wildcard result the `ns` check already computes inside `scan_domain` rather than re-probing.
- Do not score the category 0 on a wildcard zone; report it as not assessable for this signal.

PR #958, opened 2026-09-09
fix(ns,zone-hygiene): the wildcard canary sees every record family (#942)

### LINKED ISSUE #942: check_ns / check_zone_hygiene: AAAA-only (and, for check_ns, dangling-CNAME) wildcards are invisible to the A-only canary — clean 'no sensitive subdomains resolve' verdict on a wildcard zone
## Summary

Both wildcard canaries in the scanner ask for **A records only**, and the DoH record layer filters answers to the requested type. So a zone whose wildcard is AAAA-only (`*.zone AAAA …`) — and, for `check_ns`, one whose wildcard is a dangling CNAME with no address — reads as "no wildcard". Downstream, `check_zone_hygiene`'s sensitive-subdomain sweep is also A-only, so the same zone earns the clean **"No sensitive subdomains resolve publicly"** verdict for names that do resolve (over AAAA).

Score-neutral (the wildcard `medium` in `check_ns` is not emitted, the sweep flags nothing) — this is a wording/verdict defect, not a scoring one. Surfaced by the #930 investigation; PR #934 documents the blind spot rather than closing it.

- `packages/dns-checks/src/checks/check-ns.ts:301` — `queryDNS(probeFqdn, 'A', { timeout })` is the only wildcard canary in `check_ns`.
- `src/lib/dns-records.ts:41` — `queryDnsRecords` returns `Answer` entries **filtered to `RecordType[type]`**, so a CNAME-only answer to an `A` query (dangling wildcard alias) yields `[]`; an AAAA-only wildcard answers nothing to `A` at all. The package adapter (`src/lib/dns-query-adapter.ts:65`) routes non-TXT types through this function.
- `src/tools/check-zone-hygiene.ts:129` — the sensitive-name sweep is `queryDnsRecords(fqdn, 'A', dnsOptions)`; nothing resolving → the clean verdict.
- PR #934 (open) adds a raw-read canary in front of that sweep so a **CNAME-only** wildcard answer counts as a wildcard for `check_zone_hygiene`, and its CHANGELOG entry states: "Known blind spot (shared with `check_ns`): canary and sweep are A-only, so an AAAA-only wildcard is not detected." It does not touch `check_ns`.

So after #934 merges the residual is:

| Wildcard shape | `check_ns` "Wildcard DNS detected" | `check_zone_hygiene` sweep |
|---|---|---|
| `*.zone A` | detected | handled (#934) |
| `*.zone CNAME` → resolves | detected (A follows the CNAME) | handled (#934) |
| `*.zone CNAME` → dangling | **missed** (type-filtered `[]`) | handled (#934, raw read) |
| `*.zone AAAA` only | **missed** | **missed** — clean verdict on names that resolve |

## Proposed handling

- `check_ns` canary: query `A` and `AAAA` (one extra DoH query per scan; `ns` is not in the parallelism-candidates list so keep it serial) and read the answer raw (or via the same `lookupA`-style helper #934 introduces) so a CNAME-only answer also counts. Emit the same `Wildcard DNS detected` finding; metadata can note which family answered.
- `check_zone_hygiene`: either extend the canary + sweep to AAAA, or — cheaper — when the canary is AAAA-only-positive, **withhold** the clean "No sensitive subdomains resolve publicly" verdict the way #934 already does for A-wildcards, and skip the sweep. Withholding is the honest minimum; the full AAAA sweep doubles the sweep's query count for a case that is rare in practice.
- Add fixtures for `AAAA`-only and dangling-`CNAME` wildcard zones to `packages/dns-checks/src/__tests__/checks/` (ns) and `test/check-zone-hygiene*.spec.ts`.

Measure before widening: an AAAA-only wildcard with no A wildcard is uncommon; if a corpus pass shows it is <0.5% of zones, the withhold-only route is sufficient.

## Refs

- #930 (the wildcard false-positive that motivated the canary), #934 (canary PR; documents this blind spot)
- CLAUDE.md → `SCAN_DNS_CONCURRENCY` is zero-sum; `ns` is explicitly not a parallelism candidate

MadaBurns:
Closes #942.

> ⚠️ **Stacked on #956** (`fix/issue-948-transient-abstentions`) for version ordering — both PRs move `SCORING_MODEL_VERSION` and the dns-checks package, which must be sequential. GitHub will retarget this to `main` when #956 merges. **Merge #956 first.**

**Scoring change — `SCORING_MODEL_VERSION` 1.26.0 → 1.27.0.** `@blackveil/dns-checks` 1.38.0 → 1.39.0 (`PARITY_CORPUS_VERSION` in lockstep) — bv-web-prod re-vendor required.

## The defect

Both wildcard canaries queried **A only**, and the DoH record layer filters answers to the requested type. So two real wildcard shapes read as "no wildcard":

| Wildcard shape | `check_ns` before | `check_zone_hygiene` before |
|---|---|---|
| `*.zone A` | detected | handled (#934) |
| `*.zone CNAME` → resolves | detected | handled (#934) |
| `*.zone CNAME` → dangling | **missed** | handled (#934) |
| `*.zone AAAA` only | **missed** | **missed** — clean verdict on names that resolve |

## The issue's central assumption was wrong

#942 assumed the package could only see type-filtered answers, and that this layering blocked a package-side fix. It does not: `RawDNSQueryFunction` exists in the package's own types, `checkNS` **already accepts** `options.rawQueryDNS`, and the Worker always supplies it. So this reimplements the raw-read shape package-side (mirroring the existing in-file `probeNameserverReachable` precedent) rather than importing #934's Worker-side `lookupA` across the layering boundary.

## `check_ns`

One **raw** A query now settles both the A and the dangling-CNAME shape at **zero extra cost** — the CNAME rides in on the answer it was already fetching. A second AAAA query is spent **only** when that answer came back completely empty.

Serial and conditional, never a `Promise.all`: `ns` is explicitly not a parallelism candidate and `SCAN_DNS_CONCURRENCY` is zero-sum across 19 categories. Cost is **+1 subrequest on zones with no A wildcard, none on a wildcard zone**.

`rawQueryDNS` stays **optional** — a consumer that omits it (bv-web-prod calls `checkNS` from the vendored tarball) keeps the historical A-only path verbatim. Pinned by a test.

⚠️ **This half moves scores, on an unmeasured population.** A newly-detected wildcard is the *existing* `medium` at its *existing* −15, so `ns` goes 100 → 85 and the clean "Nameservers properly configured" finding is suppressed. Downward only. Prevalence of AAAA-only and dangling-CNAME wildcards is **unmeasured** precisely because the defect made those zones indistinguishable from wildcard-free ones — written up as unmeasured rather than claimed as zero.

## `check_zone_hygiene` — score-neutral by construction

An AAAA-only wildcard now yields a **distinct `detected_ipv6`** outcome: one `info` note (0 penalty) and the clean verdict withheld, which an IPv4-only sweep cannot support on such a zone.

Deliberately **not** routed through the `inconclusive` / `checkStatus: 'error'` abstention lane — the A sweep genuinely ran, and an AAAA wildcard cannot fabricate an A answer, so every IPv4 hit stays **real evidence**: it keeps its scored `medium` and its place in the "Excessive exposure" count.

The status is a **distinct arm rather than a `family` flag on `detected`** precisely so `isWildcardSynthetic` can never compare a hit's IPv4 answers against v6 addresses — which would fold a real hit into nothing and silently downgrade a genuine `medium`. Structural prevention, not convention. Pinned by a test.

A **full AAAA sweep was rejected**: ~11 extra queries, and on an AAAA wildcard zone it would manufacture ten new `medium` findings — reintroducing #930 verbatim in the v6 dimension — on a check that is `scanIncluded: false` and whose category score never reaches `computeScanScore`.

## 🚨 The test trap this nearly hid behind

`test/check-zone-hygiene.spec.ts`'s wildcard mock dispatched on `url.includes('type=A')` — and `'type=AAAA'.includes('type=A')` is **true**. Left alone, every new AAAA query would have been answered from the A arm: the tests would have passed for the wrong reason while the fix did nothing. A textbook fail-open measurement.

The dispatch now extracts and compares the type exactly, and the new test asserts a **positive control** on the AAAA query count — if the A arm swallowed it, that count is zero and the test fails.

## Verification

- `Tests 142 passed` (zone-hygiene + ns + abstention audit), `18 passed` (new package canary spec)
- Extra regression sweep, unrequested: all of `packages/dns-checks/src/__tests__/checks/` → `503 passed`; 13 specs matching `checkNs|_bv-probe-|zone_hygiene` → `190 passed`; chaos suite green
- `npm run typecheck` RC=0; `typecheck:tests` ratchet `TOTAL: 735 (baseline 735, delta +0)`
- Version-lockstep contracts re-verified after the bump: `93 passed`

Three `prettier --check` warnings on touched files are **pre-existing on main** (verified by piping `git show HEAD:<file>` through prettier); every delta sits in untouched legacy code, so they were left rather than dragging unrelated reformatting into the diff.


MadaBurns:
## Adversarial review — no defects found; one follow-up noted, deliberately not actioned

### The trap is genuinely closed — verified, and the test proven non-cosmetic

The headline risk was that `test/check-zone-hygiene.spec.ts`'s mock dispatched on `url.includes('type=A')`, where `'type=AAAA'.includes('type=A')` is **true** — every AAAA query answered from the A arm, tests passing for the wrong reason while the fix did nothing.

Confirmed closed: the dispatch now extracts the query-string token via `url.match(/[?&]type=([^&]+)/)?.[1]` and compares **exactly**, with an in-file comment naming the trap. The reviewer reasoned through a manual revert and confirmed the new test genuinely fails without the fix: with an A-only canary, zero AAAA fetches are issued, so `canaryNames(aaaaQueries).size` is `0 ≠ 1`.

### Verified against the code, not the comments

- **`isWildcardSynthetic` can never see v6 addresses** — and it is enforced by the *type system*, not convention. `WildcardProbe` is a discriminated union, and `isWildcardSynthetic` is only called inside `status === 'detected'` branches, so TypeScript narrowing makes a `detected_ipv6` value structurally unable to reach it. That was the point of using a distinct status rather than a `family` flag: folding a real IPv4 hit against v6 wildcard addresses would silently downgrade a genuine `medium`.
- **Score-neutrality checked against the real penalty table**, not the claim: `SEVERITY_PENALTIES.info = 0` in `packages/dns-checks/src/types.ts`, and the new finding is `info` with no `missingControl`.
- **The AAAA probe is serial-conditional** (`else if (aAnswers.length === 0)`, sequential `await`, never `Promise.all`) — and the reviewer independently confirmed against `tool-definitions.ts` that `check_ns` is `scanIncluded: true` (so the conditionality genuinely matters for scan budget) while `check_zone_hygiene` is `scanIncluded: false` (so its extra query is direct-call latency only, never part of the 19-category fan-out). That verified the PR's claim against the registry rather than the prose.
- **`rawQueryDNS` fallback is safe** — the `else` branch calls the original type-filtered `queryDNS` with no raw-shape dereference, pinned by a dedicated test asserting both the detected/absent shapes and the query count.
- **No new CNAME false-positive surface**: the `'cname'` branch fires only when the A answer holds a CNAME *and no A record*, a strict subset of the pre-existing accepted risk; `dnssecFlag=false` is passed explicitly, so DNSSEC records cannot leak in.
- **The three re-expressed budget assertions were strengthened, not loosened** — each gained `expect(aaaaQueries).toEqual([])`.

### Noted, not actioned

One *untouched* adjacent test (`keeps the non-wildcard sweep byte-identical apart from the single canary query`) still uses its own inline mock with the old substring dispatch, and asserts on distinct probe **names** rather than fetch **calls**. The reviewer traced it and confirmed it masks no functional bug — the mock's fallthrough returns an empty `Answer` for both the A and misrouted AAAA query, so the assertion holds either way.

Leaving it: migrating it to the shared `wildcardMock()` helper is a test-quality improvement to a file this PR already touches heavily, and folding it in would mean rewriting a test whose current pass is genuinely correct. Worth doing, but as its own change — not smuggled into a scoring PR.

*Caveat:* this reviewer had no shell access, so the above is verified by reading the raw branch contents and tracing dependent source, not by execution. The specs and build were run separately: `142 passed`, `18 passed` on the new package spec, plus a `503 passed` regression sweep across the package check suite.

Issue #643, opened 2026-08-07
Non-mail post-processing is DEAD on real DNS data: hasNoMx matches a finding title nothing emits


MadaBurns:
Found while fixing the MTA-STS false-text bug. **This is the root cause of that bug, and it is much broader than the symptom.**

## The defect

Two sites gate non-mail handling on a finding title that **no production code path emits**:

```
src/tools/scan/post-processing.ts:80
  const hasNoMx = mxResult.findings.some((f) => f.title === 'No MX records found')

src/tools/scan/maturity-staging.ts:306
  const hasNoMx = measured(mxCheck) && mxCheck.findings.some((f) => f.title === 'No MX records found')
```

`check-mx` never emits that title. Its actual non-mail titles are:

- `Null MX record (RFC 7505)`
- `Correctly-configured non-mail domain`
- `Non-mail domain SPF not hard-fail`
- `No MX and no SPF — domain spoofable`
- `MX records not assessed`
- `Dangling MX record`

Verified by grep: the string `'No MX records found'` occurs **only at the two comparison sites above** — never as an emitted value.

**So `hasNoMx` is always `false` against real DNS.**

## What that silently disabled

1. **`adjustForNonMailDomain()` never runs.** CLAUDE.md documents this behaviour as live:
   > *"Non-mail (no MX): parent DMARC `sp=`/`p=` → downgrade email-auth findings to `info`"*

   It is gated entirely on `hasNoMx`. Non-mail domains have therefore been carrying **un-downgraded critical/high email-auth findings** that the scoring model intends to be `info` — i.e. systematically over-penalised.

2. **`adjustBimiForNonMailDomain()` never runs** — it was only called inside the dead branch. (Symptom now fixed by calling it on the correct condition.)

3. **`clarifyMtaStsForMailDomain()` ran on every domain**, including null-MX ones, overwriting `check-mta-sts`'s already-correct non-mail copy with "Since this domain has MX records and accepts email…". This was the customer-visible falsehood that led here.

4. **Maturity staging** applies the same dead predicate at `maturity-staging.ts:306`, so non-mail maturity is computed on the mail-domain path.

## Status

The **symptoms** (3) and (2) are fixed by keying on a real signal — a new `mxDeclaresNoInboundMail()` predicate reading `controlPresent` (which `check-mx` sets `false` on both the no-MX and null-MX terminal paths, `true` only for real mail routing, `undefined` when inconclusive), with the phantom title retained only as a legacy fallback.

[...]

PR #959, opened 2026-09-09
fix(mx): name a loopback MX instead of mislabelling it (#944)

### LINKED ISSUE #944: Decide: is a '10 localhost' MX a no-mail signal? isNullMxRecord matches only RFC 7505 '0 .'; check_mx and map_supply_chain treat localhost as a live exchange
## Summary

Decide-and-document. #932's suggested-fix list mentioned `localhost` alongside `0 .` as a "no inbound mail" MX idiom. The shared classifier `isNullMxRecord` (made the single implementation in #933) matches only the RFC 7505 form — `exchange === '' || exchange === '.'`. An MX of `10 localhost` (or `localhost.`) is therefore treated as a live exchange everywhere:

- `check_mx` counts it as a real mail exchange (presence finding, single-MX finding, MX-to-IP check all run against it; nothing resolves so the A/AAAA follow-ups produce their usual "MX host does not resolve" style findings).
- `map_supply_chain` renders `localhost` as an email-receiving provider row.
- `scan_domain`'s non-mail post-processing keys on `controlPresent: false` from `check_mx`'s no-mail terminal paths, so a `localhost` MX domain is treated as **mail-enabled** — DKIM / MTA-STS / BIMI "missing" findings keep their mail-enabled severities instead of the no-send `info` downgrade.

The question is whether `localhost` is a common enough idiom to classify, and if so what it should mean.

- `packages/dns-checks/src/checks/mx-analysis.ts:36-38` — `isNullMxRecord(record: Pick<ParsedMxRecord, 'exchange'>)` returns `record.exchange === '' || record.exchange === '.'`.
- `src/tools/map-supply-chain.ts:357-358` — `hasNullMx = rawMxRecords.some(isNullMxRecord)`; every non-null exchange becomes a provider host.
- `src/tools/scan/post-processing.ts:508-530` — non-mail recognition is structural (`controlPresent`), with `NO_INBOUND_MAIL_MX_TITLES` as a legacy fallback; neither path has a `localhost` case. `grep -rni localhost packages/dns-checks/src/checks/check-mx.ts packages/dns-checks/src/checks/mx-analysis.ts src/tools/check-mx.ts src/tools/map-supply-chain.ts src/tools/scan/post-processing.ts` → no matches.
- #932 issue body, suggested fix: "Null MX (`0 .` / `localhost`): emit no email-receiving dependency". #933 implemented the `0 .` half.

## Proposed handling

Two defensible positions — pick one and record it in the `isNullMxRecord` JSDoc so the question is not re-raised:

**A. Do not classify (status quo, documented).** `localhost` is not a standard; RFC 7505 is explicit that null MX is `0 .`. A `localhost` MX is a misconfiguration (mail to the domain is undeliverable *and* some senders will try 127.0.0.1) and the honest output is a finding saying so, not a silent reclassification to "non-mail". Under this option the follow-up is a dedicated `medium`/`low` finding in `check_mx` ("MX points at localhost — undeliverable; publish `0 .` per RFC 7505 if no mail is intended") and `map_supply_chain` dropping loopback exchanges from provider rows the same way it drops the null MX.

**B. Classify as a no-mail signal.** Treat `localhost`, `localhost.`, and loopback literals (`127.0.0.1`, `::1`) as equivalent to null MX for the *no-inbound-mail* decision, with an `info` note recommending `0 .`. Cheaper for users who inherited the idiom, but it silently rewards a non-standard config and changes scan_domain post-processing (a domain would flip to the non-mail wording and severities), which is a score-affecting change on any domain using it — needs a corpus count before shipping.

Suggested first step for either: a corpus measurement of how many zones in the 1000-domain scoring corpus publish an MX exchange of `localhost`/loopback. If it is effectively zero, take A (finding + supply-chain filter) and close.

## Refs

- #932 (suggested-fix list), #933 (shared `isNullMxRecord`, RFC 7505 form only)
- CLAUDE.md → "Post-processing: Non-mail (no MX) downgrades … only under inherited DMARC enforcement" — the load-bearing gate that any option B would have to respect

MadaBurns:
Closes #944.

> ⚠️ **Stacked on #958** (which is stacked on #956) for version ordering — all three move `SCORING_MODEL_VERSION` and the dns-checks package, which must be sequential. **Merge order: #956 → #958 → this.** GitHub retargets automatically as each base merges.

**Scoring change — `SCORING_MODEL_VERSION` 1.27.0 → 1.28.0.** `@blackveil/dns-checks` 1.39.0 → 1.40.0 (`PARITY_CORPUS_VERSION` in lockstep) — bv-web-prod re-vendor required.

## The defect

An MX of `0 localhost.` was treated as a live mail exchange everywhere. Confirmed live against a real control domain:

- `check_mx` → generic `medium` **"Dangling MX record"**, which only fires *at all* because Cloudflare and Google both NXDOMAIN `localhost`. RFC 6761 permits a recursive resolver to answer it — on such a resolver the domain scored **95 with no finding whatsoever**.
- `map_supply_chain` → a `trustLevel: critical` **`localhost` email-receiving provider row**.

## The decision, and what settled it

This was a decide-and-document issue. **Option A — report it as the defect it is; do not reclassify it as no-mail** — chosen on measured prevalence, not preference.

| Sample | Loopback MX |
|---|---|
| Rebuilt stratified Tranco-1000 corpus | **0 / 992 measured (0.00%)** |
| Wider 29,780-domain sample | **15 / 29,385 (0.051%)**, ~1 in 1,960 |

The zero is trustworthy because **the positive control passes**: the same script found those 15 real zones, and one was confirmed from three independent resolver vantages. A bare zero would have been indistinguishable from a broken query.

**The shape is what settled it:** all 15 wild hits are the *identical* string `0 localhost.` — priority 0, exchange `localhost`, zero IPv4-loopback and zero `::1` forms. That is an operator reaching for RFC 7505 `0 .` and getting the exchange wrong. A finding that says so is strictly more useful than a silent reclassification to "non-mail" that hides the mistake *and* re-grades the domain.

`isNullMxRecord` stays RFC-7505-only, with the decision recorded in its JSDoc so the question is not re-raised. `controlPresent` stays `true` and `scan_domain`'s non-mail downgrade does not fire — `src/tools/scan/post-processing.ts` is deliberately **unchanged**, which is the point of Option A, and a negative test in the #643 block pins it so a future Option-B attempt cannot land silently.

## Score-neutral by construction — measured, not asserted

Loopback records are excluded from the IP-target **and** dangling-MX passes, so the new `medium` **replaces** the one those would have emitted rather than stacking.

```
CONTROL   score=80  [info:MX records found, medium:Dangling MX record,   low:Single MX record]
LOOPBACK  score=80  [info:MX records found, medium:MX points at localhost, low:Single MX record]

SHIPPED (replaces) = 80 | if it stacked with dangling = 65 | with a third medium = 50
```

`mx` has **no severity cap**, so stacking here is a real hazard, not theoretical. Those two exclusions are the whole safety of the change and ship together or not at all.

## ⚠️ The matcher is loopback-only, never "unroutable"

22 domains in the sample published Microsoft 365 `msNNNNNNNN.msv1.invalid` verification pseudo-MX on **healthy** tenants. An unroutable-target matcher would hand every ordinary M365 customer a `medium` — the same over-match class as the TXT-hygiene `SERVICE_SPF_DOMAINS` incident. The `.localhost` suffix is anchored on a label boundary so `localhostings.com` does not match, and the rendered exchange list is capped since DNS data is caller-controlled.

`check_mta_sts`'s `detectInboundMail` is left **unchanged**, with a comment recording why: under Option A we do not reclassify localhost as no-mail, so MTA-STS correctly keeps treating such a domain as inbound-mail-receiving. That seam is the one place Option A reads slightly oddly, so it is documented rather than left for someone to trip on.

## Why a version bump if it is score-neutral

It is a **new detection family**, and it makes the verdict **resolver-independent** — which moves the score on any resolver that honours RFC 6761 for `localhost`. Per the release skill, `packages/dns-checks/src/checks/` changes are *not* caught by the blocking contract gate, so the package + corpus bump is done by hand.

## Tests

`Tests 237 passed` across the five scoped files; `npm run typecheck` RC=0; parity-corpus contract green after the bump. Includes the suffix-collision guard (`localhostings.com`), the `.invalid` negative, both anti-stacking assertions, the supply-chain provider-row removal, and the post-processing negative case.


MadaBurns:
## Adversarial review — completed, one residual adjudicated

An independent review traced the anti-stacking math by hand across all four scenarios (single `localhost`, single `127.0.0.1`, multiple loopback records, mixed real+loopback), the `.invalid`/`.localhost` false-positive guards, normalisation, detail sanitisation, the supply-chain signal severity, `scan_domain` post-processing isolation, and the three-version lockstep. **No blocking findings.**

Two things worth recording from it:

**1. A normalisation detail that is load-bearing, not redundant.** `isLoopbackMxRecord` re-lowercases and re-strips the trailing dot even though `parseMxRecords` already does. That is *not* dead code: the Worker-side `queryMxRecords` (`src/lib/dns-records.ts`) does **not** lowercase, and `map_supply_chain` calls through that path. Removing the internal normalisation would break loopback detection for supply-chain callers only — a nasty half-working failure.

**2. One genuine residual, adjudicated and documented rather than fixed.** The reviewer noticed `getSingleMxFinding()` is called with unfiltered `mxRecords` while the IP-target and dangling passes use `routableRecords`, and asked whether that is an oversight.

It is not — and changing it would have been a **silent scoring change**:

- The measured population is a lone `0 localhost.`. Filtering leaves **zero** routable records → `getSingleMxFinding` returns null on `length !== 1` → the domain scores **85 instead of 80**.
- That is a leniency change wearing the costume of a cleanup, and it would break the score-neutrality this PR is built on.
- The behaviour is also **pre-existing on `main`**, so leaving it is preservation, not oversight.

The reviewer did find a real (also pre-existing) false negative underneath it: a zone publishing one real exchange **beside** a loopback one has no actual redundancy but escapes the finding, because the raw count is 2. That is a scoring change in its own right and belongs in its own PR with its own version bump — not smuggled in beside a false-positive fix.

Both are now recorded in a comment at the call site so a future reader cannot "tidy" it into a regression (6abd546). Comment only; no behaviour change; `24 passed` on the scoped specs after.

*Caveat on review coverage:* this reviewer had no shell access, so its conclusions come from static tracing rather than execution. The anti-stacking property itself was separately confirmed by running the tests — score 80 on both the `localhost` and `127.0.0.1` cases, versus 65 if the finding stacked with dangling-MX.

PR #989, opened 2026-09-13
fix(dns-checks): read the SPF `all` qualifier from a term on the finding path too (#988)

### LINKED ISSUE #988: check_spf: the `all` qualifier is matched as a substring — a hostname containing `-all` suppresses the soft-fail finding
## The defect

`packages/dns-checks/src/checks/check-spf.ts:173` extracts the `all` qualifier with an **unanchored, first-match** substring scan:

```ts
const allMechanism = spf.match(/[+?~-]all/i);
```

Any hostname containing `-all` / `+all` / `~all` **before** the terminal mechanism wins that match — `send-all.`, `mail-all.`, `smtp-all.` are ordinary names. The captured value then drives the finding branches (`RISKY_MECHANISMS` check, then `=== '~all'`).

## Impact: a soft-fail record silently gets no soft-fail finding

```
v=spf1 include:send-all.example.net ~all
```

`allMechanism[0]` is `-all` (from inside the hostname). `-all` is not in `RISKY_MECHANISMS` and is not `~all`, so **every** finding branch is skipped: no "SPF soft fail (~all)" finding, no "SPF soft fail (~all) with DMARC enforcement". The domain scores as though it published hard-fail.

The inverse is also reachable: a record whose hostname contains `+all` or `?all` would emit a **critical** "Permissive SPF" finding against a domain that actually publishes `-all` — a false adverse claim.

Second, smaller defect at the same site: a **bare `all`** is not matched at all, because the character class requires a qualifier. RFC 7208 §4.6.2 gives an unqualified mechanism the implicit `+`, so `v=spf1 include:… all` is the *most permissive* posture and currently produces no permissive-SPF finding.

## Why this is filed rather than fixed in the PR that found it

PR #987 adds a structured `spfAll` signal and hit this bug directly — its first implementation reused this regex and reported `-all` for the record above. That PR fixed the extraction **for its own signal only** (splitting into RFC 7208 terms and matching `^[+?~-]?all$` on whole terms), and deliberately left this finding path alone, because:

- the finding severities are **score-bearing**, so correcting them moves scores, and
- PR #987 guarantees score neutrality and asserts it against the unedited scoring-contract test and parity fixtures.

So the two paths now diverge on purpose, documented at the call site. This issue is the other half.

## Suggested fix

Reuse the same term-based extraction for the finding path, then re-baseline whatever scores move. Expect real movement: domains currently escaping a soft-fail finding will start receiving one, and any domain wrongly flagged critical will stop. That is a scoring change and wants its own release note and a `SCORING_MODEL_VERSION` decision.

## Reproduction

```
node --input-type=module -e "
import { checkSPF } from './packages/dns-checks/dist/index.js';
const dns = (r) => async (d) => r[d] ?? [];
const res = await checkSPF('example.com', dns({'example.com':['v=spf1 include:send-all.example.net ~all']}));
console.log(res.score, res.findings.map(f => f.title));
"
```
Observed: no soft-fail finding. A control record (`v=spf1 include:_spf.example.net ~all`) does produce one.

Verified against `origin/main @ 5dc0846fb`; the regex is unchanged at that base, so this is pre-existing and not introduced by #987.

Possibly related: #909 (the `~all` recommendation being inverted across the DMARC branches) touches the same branch structure.

MadaBurns:
Closes #988.

## The defect

`check_spf`'s FINDING path extracted the `all` qualifier with an unanchored, first-match substring scan — `spf.match(/[+?~-]all/i)` at `packages/dns-checks/src/checks/check-spf.ts:173` on the base commit (`git show 7dce66041:packages/dns-checks/src/checks/check-spf.ts | sed -n '173p'`).

Any hostname containing `-all` / `+all` / `?all` **before** the terminal mechanism wins that match. `send-all.`, `mail-all.`, `smtp-all.` are ordinary names. Measured on the unfixed source (`npx tsx` against `src/checks/check-spf.ts`, mock DNS, no DMARC record):

| record | old finding | old spf score |
|---|---|---|
| `v=spf1 include:send-all.example.net ~all` | *(none)* | 100 |
| `v=spf1 include:send+all.example.net -all` | critical `Permissive SPF: +all` | 60 |
| `v=spf1 include:send-all.example.net` | *(none)* | 100 |
| `v=spf1 include:_spf.example.net all` | medium `No 'all' mechanism` | 85 |

Row 1 is a soft-fail record scoring as hard fail. Row 2 is a **false adverse claim** at critical severity against a domain that publishes `-all`. Row 3 is a record with no `all` term reading as clean. Row 4 is RFC 7208 §4.6.2's implicit `+` — the most permissive posture there is — reported as merely unspecified.

#987 hit exactly this bug for its new `spfAll` metadata signal and fixed it there with a term-based extraction, deliberately leaving the finding path alone because the finding severities are score-bearing and #987 guaranteed score neutrality. That divergence is documented at the call site in `git show 7dce66041 -- packages/dns-checks/src/checks/check-spf.ts`. This PR is the other half.

## The fix — one extraction, both consumers

The #987 term split is hoisted to `allTerm` / `allQualifier` and now drives **both** the finding branches and `metadata.spfAll`. No second parser was written; the two paths are literally the same expression, so they cannot diverge again (asserted by the last test in the new file, which checks `metadata.spfAll` and the finding agree on the decoy shape).

Rules, per RFC 7208: terms are whitespace-separated (§4.6.1); the first whole term matching `^[+?~-]?all$` wins, because `all` always matches so evaluation never reaches a later one (§5.1); a bare `all` normalises to `+all` (§4.6.2).

## ⚠️ THIS CHANGES SCORES — it is not score-neutral

Measured before and after on the same inputs with the same harness (spf category score from `checkSPF`, mock DNS, no DMARC record):

| record | score before | score after | finding delta |
|---|---|---|---|
| `v=spf1 include:send-all.example.net ~all` | 100 | 95 | **+** low `SPF soft fail (~all)` |
| `v=spf1 include:send+all.example.net -all` | 60 | 100 | **−** critical `Permissive SPF: +all` |
| `v=spf1 include:_spf.example.net all` | 85 | 60 | **−** medium `No 'all' mechanism`, **+** critical `Permissive SPF: +all` |
| `v=spf1 include:send-all.example.net +all exp=why.example.net` | 100 | 60 | **+** critical `Permissive SPF: +all` |
| `v=spf1 include:send-all.example.net` | 100 | 85 | **+** medium `No 'all' mechanism` |
| `v=spf1 ~all -all` | 95 | 95 | unchanged (first term wins; both extractions agree) |
| `v=spf1 include:_spf.example.net -all` | 100 | 100 | unchanged (control) |
| `v=spf1 include:_spf.example.net ~all` | 95 | 95 | unchanged (control) |
| `v=spf1 +all` | 60 | 60 | unchanged (control) |

Both directions move: domains escaping a soft-fail or missing-`all` finding start receiving one, and a domain wrongly called critical stops being called that.

**Scope of the movement, stated precisely:** only records where the OLD substring scan and the NEW term extraction disagree change. That is (a) records containing an `all` substring inside a term that is not an `all` mechanism, and (b) records whose `all` mechanism is bare. Every other record is byte-for-byte identical — evidenced by the package suite and the scoring contract and parity fixtures passing **unedited**.

**`SCORING_MODEL_VERSION` is unchanged.** The scoring model did not move; the extraction feeding it did. I did not re-baseline any corpus and make no claim about aggregate corpus impact — I did not measure one.

**Downstream, not edited here:** `src/tools/assess-spoofability.ts:135-137` derives its spf disposition from these exact finding titles (`Permissive SPF:`, `SPF soft fail`, `No 'all' mechanism`), so it inherits the corrected verdicts.

## Version lockstep

`@blackveil/dns-checks` 1.42.0 → 1.43.0 and `PARITY_CORPUS_VERSION` 1.42.0 → 1.43.0, in the same commit, as `check:scoring-contract-change` requires (the gate's `BEHAVIOR_PATHS` includes `src/parity-fixtures.ts`, which this diff touches). `SCORING_CONTRACT.packageVersion` derives from `PARITY_CORPUS_VERSION` and needs no separate edit. `CHANGELOG.md` is untouched — it tracks the root server release line.

MINOR, not patch: this changes observable check behaviour for a class of records, and downstream consumers pin this package by version (bv-web-prod vendors it as an exact tarball, currently 1.36.0).

## Verification — every command re-runnable from the repo root

**Tests written FIRST, observed RED** (`npx vitest run src/__tests__/checks/spf-all-term-findings.test.ts` from `packages/dns-checks`, before the `check-spf.ts` edit): `Tests  6 failed | 2 passed (8)`. After the fix: `Tests  8 passed (8)`.

```
cd packages/dns-checks && npx vitest run
  ->  Test Files  63 passed (63)
      Tests  1152 passed (1152)

cd packages/dns-checks && npx tsc --noEmit          ->  rc=0 (no output)
npm run typecheck                                    ->  rc=0
npm run lint                                         ->  rc=0
npx vitest run test/check-spf.spec.ts test/scoring-determinism.spec.ts
  ->  Test Files  2 passed (2)   Tests  34 passed (34)

SCORING_CONTRACT_BASE_REF=7dce66041 npm run check:scoring-contract-change
  ->  scoring-contract change gate: OK   (rc=0)
```

The gate was proven to **discriminate** rather than fail open: with `packages/dns-checks/package.json` temporarily reverted to `1.42.0` and everything else unchanged, the same command reports

```
scoring-contract change gate: BLOCKED
- scoring behavior changed while @blackveil/dns-checks remained at 1.42.0
- bump packages/dns-checks/package.json and PARITY_CORPUS_VERSION in the same PR
```

with `rc=1`. The gate fires on this diff and is satisfied only by the lockstep bump; it was not suppressed or routed around.

## Not verified

- `packages/dns-checks/src/parity-fixtures.ts` fails `npx prettier --check`. That is **pre-existing** — it fails identically at `7dce66041` with this PR's change stashed — and is not corrected here, being outside this PR's scope.
- No corpus-wide measurement of how many real domains move. I have no such measurement and make no such claim.
- Full root suite result is reported in a follow-up comment rather than asserted here.


MadaBurns:
Root suite result, as the PR body said would be reported here.

`npx vitest run` from the repo root in this branch's worktree:

```
 Test Files  1 failed | 728 passed | 4 skipped (733)
      Tests  9429 passed | 14 skipped (9443)
```

Zero test failures. The one failed **suite** is a load error, not an assertion:

```
FAIL  |workers| test/wasm-integration.test.ts [ test/wasm-integration.test.ts ]
Error: No such module "…/.worktrees/crates/bv-wasm-core/pkg/bv_wasm_core_bg.wasm?mf_vitest_force=CompiledWasm".
```

Environmental, not this diff. `crates/bv-wasm-core/pkg` is a build artifact that exists in the main checkout but was never built in this worktree — `ls -d crates/bv-wasm-core/pkg` succeeds in the main checkout and returns "No such file or directory" in the worktree. This diff touches no wasm, no Rust, and no test/ file. CI builds it, so this suite is expected to pass there; if it does not, that is a separate pre-existing issue and not caused by this change.

PR #990, opened 2026-09-13
feat(dns-checks): NZ SGE per-domain compliance evaluator (Phase 1)

MadaBurns:
## What this is

A **per-domain** NZ Secure Government Email (SGE) evaluator inside `@blackveil/dns-checks`. It answers one question — *"is THIS domain compliant with NZ SGE?"* — across the six controls an agency is measured on: DMARC `p=reject`, SPF `-all`, DKIM, SMTP TLS, MTA-STS `mode: enforce`, TLS-RPT. It is not an aggregate statistic and not a compliance-framework map.

Phase 1 only: no MCP tool, no `map_compliance` rows, no check changes, no scoring changes, no version bump.

## Public API

```ts
import { evaluateSgeCompliance, SGE_CONTROL_IDS } from '@blackveil/dns-checks';

evaluateSgeCompliance(domain: string, results: readonly CheckResult[], options?: { smtpTls?: 'enforced' | 'not_enforced' }): SgeEvaluation
```

`SgeEvaluation` = `{ domain, verdict, mailTransport, controls, counts }`; each of the six `controls` carries `status`, an optional `notMeasuredReason`, and an `evidence` array naming the exact signal read and its value.

Verify the surface:
```
git show 60af806e5 --stat
git diff 7dce66041..60af806e5 -- packages/dns-checks/src/index.ts
```

## Three-state, never a boolean

`satisfied` / `not_satisfied` / `not_measured`, with a typed reason on every unmeasured control (`check_absent`, `check_not_completed`, `signal_absent`, `policy_unreadable`, `selector_enumeration_inconclusive`, `no_mail_exchanger`, `no_transport_probe`). A boolean compiles UNKNOWN silently into `false`, which on a security control is an affirmative adverse claim from zero evidence. All six controls always appear in the output. `compliant` requires all six `satisfied` — a single unmeasured control yields `indeterminate`, never a pass.

## Oracles it refuses

- **`passed`** — means "did not penalize", not "exists" (#705/#706).
- **`controlPresent` as a generic presence oracle** — it is enforcement-shaped; a published `p=none` reads `false`. Read only where its enforcement meaning *is* the question (dmarc "enforcing?", dkim "active key?", mx "real mail?").
- **Any score, band or grade.**
- **Finding TEXT.** The one prose-derived predicate used at all is the package's own canonical `findingsIndicateMissingControl`, and only in the NEGATIVE direction (no SPF record ⇒ not `-all`), never to affirm compliance.

Measurement status is gated through the package's own `isCheckMeasured` (allowlist semantics), so an out-of-union `checkStatus` from a version-skewed cache read is unmeasured, not measured. Covered by a test.

SPF and MTA-STS go exclusively through the 1.42.0 structured readers `spfAllQualifier()` / `mtaStsPolicyMode()` added by #987. The `check_spf` FINDING path still matches `all` as an unanchored substring (#988), so an `include:` hostname containing `-all` reads as a hard fail in the findings — there is a regression test asserting the evaluator follows the reader, not the prose.

## Two deliberate decisions

**DKIM selector miss is `not_measured`, not `not_satisfied`.** DKIM has no discovery mechanism; a domain signing with an uncommon selector is indistinguishable from one that does not sign. DKIM is also one of the five categories that emit no `recordPresent` (the #705 residual gap), so there is no fallback oracle.

**No MX is not non-compliance.** A domain with no MX (or a null MX) receives no mail, so the three inbound-transport controls are excused to `not_measured` / `no_mail_exchanger` rather than failed — failing a domain for not securing mail it cannot receive is an affirmative adverse claim about a posture NIST SP 800-177r1 recommends. The three anti-spoofing controls are still evaluated: MX is inbound, spoofing is not. Stated consequence: a no-MX domain can never reach `compliant`, only `indeterminate`. Four tests cover it (`describe('a domain with no MX')` in `sge-evaluate.test.ts`).

## Known limitation, recorded not papered over

`check-tlsrpt` sets `recordPresent: true` for a published record carrying no `rua=` — a record that names no reporting address. No structured signal separates the two states, and inferring it from the finding title is forbidden, so the evaluator reports such a domain `satisfied`. Closing it needs the check to emit a structured signal the way #987 did for `spfAll` / `mtaStsMode`. Out of scope here; documented at the call site in `evaluate.ts`.

## Harness discrimination — known-good vs known-bad, both live-measured

Both run through the real check functions at this commit against live DNS/HTTPS on 2026-09-13 (DoH `cloudflare-dns.com`, real `fetch` for the MTA-STS policy file):

| control | `health.govt.nz` (good) | `stuff.co.nz` (bad) |
|---|---|---|
| DMARC p=reject | **satisfied** (`controlPresent true`, `partialEnforcement false`) | satisfied |
| SPF -all | **satisfied** (`spfAllQualifier() = '-all'`) | **not_satisfied** (`'~all'`) |
| DKIM | **satisfied** (`controlPresent true`) | satisfied |
| SMTP TLS | not_measured / `no_transport_probe` | not_measured / `no_transport_probe` |
| MTA-STS enforce | **satisfied** (`mtaStsPolicyMode() = 'enforce'`) | **not_satisfied** (`recordPresent false`) |
| TLS-RPT | **satisfied** (`recordPresent true`) | **not_satisfied** (`recordPresent false`) |
| verdict | `indeterminate` (5 satisfied / 0 failed / 1 unmeasured) | `non_compliant` (2 / 3 / 1) |

The instrument discriminates on four of the six legs, and nothing returns `compliant` on DNS evidence alone — SMTP TLS is structurally unmeasurable here, which is the honest answer rather than a silent pass.

Re-runnable check of the underlying records:
```
dig +short TXT _dmarc.health.govt.nz @1.1.1.1     # v=DMARC1; p=reject; sp=none; pct=100; rua=…
dig +short TXT health.govt.nz @1.1.1.1 | grep spf1 # … -all
curl -sS https://mta-sts.health.govt.nz/.well-known/mta-sts.txt   # mode: enforce
dig +short TXT _smtp._tls.health.govt.nz @1.1.1.1  # v=TLSRPTv1; rua=…
dig +short TXT _dmarc.stuff.co.nz @1.1.1.1
dig +short TXT stuff.co.nz @1.1.1.1 | grep spf1     # … ~all
dig +short TXT _mta-sts.stuff.co.nz @1.1.1.1        # empty
dig +short TXT _smtp._tls.stuff.co.nz @1.1.1.1      # empty
```

## Regression case — `health.govt.nz`

`src/__tests__/sge/fixtures/health-govt-nz-2026-09-13.json` is the verbatim JSON of the six `CheckResult`s this package's own check functions produced against live DNS at this commit. The test pins the **trap as well as the verdict**, so it cannot decay into a fixture that no longer discriminates:

- **Formerly a FALSE NEGATIVE.** `p=reject` with `sp=none` scores **50** and raises a **HIGH** finding ("Subdomain policy weaker than parent policy"), so every score-band and severity-count oracle reads this domain as failing DMARC. It is not: the SGE control is the organisational domain's own `p=`, which is `reject`. The evaluator reports **satisfied**.
- **DKIM.** On the live measurement this domain **does** publish an active key (`selector1._domainkey` CNAMEs to `selector1-health-govt-nz._domainkey.mohgovtnz.onmicrosoft.com`, which serves a valid `v=DKIM1` RSA key), so `satisfied` is the correct answer and the evaluator gives it — only because `controlPresent === true` was affirmatively observed. The false-positive guard is a separate test that takes the **same fixture**, flips only DKIM's structural signal to "no selector answered" while leaving `passed: true` and a pass-band score in place — exactly the shape the old oracle read as compliant — and asserts the result is `not_measured`, and neither `satisfied` nor `not_satisfied`.

⚠️ **Claim I am withdrawing:** the task premise said the old oracle produced a false POSITIVE on `health.govt.nz`'s DKIM. I could not reproduce that as a *wrong verdict* — the domain genuinely has DKIM, measured live above, so a "pass" there is correct. What is real is the `passed`-shaped oracle itself (the #705 residual gap: `spf`/`dkim`/`ssl`/`ns`/`http_security` emit no `recordPresent`), and that is what the guard test pins.

**Mutation proof.** Mutating the DMARC leg to a score-band oracle (`score >= 85`) and the DKIM leg to a `passed` oracle turns **8** tests RED across both files, including both regression assertions — observed, then reverted:
```
× DMARC p=reject is SATISFIED despite a score of 50 and a HIGH finding
× would NOT report DKIM satisfied if the key evidence disappeared but `passed` stayed true
× is indeterminate on DNS evidence alone, with SMTP TLS the single unmeasured control
× reaches compliant once a transport probe supplies the sixth control
× p=quarantine (structural partialEnforcement) is not satisfied
× ignores `passed: false` + score 0 when the structural signal says enforcing at reject
× a selector-enumeration miss is not_measured — neither satisfied nor not_satisfied
× never reads `passed: true` on a dkim result as evidence the control exists
Tests  8 failed | 41 passed (49)
```

## Verification — every command re-runnable from this branch

| check | command | result |
|---|---|---|
| new suite | `npx vitest run --root packages/dns-checks src/__tests__/sge` | **2 files, 49 passed** |
| package suite | `cd packages/dns-checks && npx vitest run` | **64 files, 1193 passed**, rc=0 |
| package typecheck | `cd packages/dns-checks && npx tsc --noEmit` | rc=0 |
| root typecheck | `npm run typecheck` | rc=0 |
| lint | `npm run lint` | rc=0 |
| scoring-contract gate | `npx tsx scripts/scoring-contract-change.ts` | rc=0 — this diff touches none of its three trigger paths (`scripts/scoring-contract-change.ts:10-12`) |
| OSS safety gate | `npm run audit:oss-safety` | rc=0 |
| root suite | `npm test` | **9470 passed, 14 skipped; 1 suite failed** — see below |

**The one root-suite failure is environmental and pre-existing.** `test/wasm-integration.test.ts` fails with `No such module ".worktrees/crates/bv-wasm-core/pkg/bv_wasm_core_bg.wasm"`. That artifact is a build product that exists in the main checkout (`ls /Applications/Github/bv-mcp/crates/bv-wasm-core/pkg`) and not in a fresh worktree; the resolved path also escapes the worktree root. This diff touches no WASM code. I did **not** re-run the root suite on an unmodified base to confirm, so this is diagnosed, not baseline-proven.

## Score neutrality

No check, no scoring module, no fixture and no version was touched. Nothing in `src/sge/` is read by the scoring path. Diff is additive only:
```
git diff --stat 7dce66041..60af806e5
# 7 files changed, 1308 insertions(+)   — 0 deletions
```

## ⚠️ Needs a version bump — deliberately not done here

A consumer cannot pin the new export without a `@blackveil/dns-checks` MINOR bump (1.42.0 → 1.43.0) plus `PARITY_CORPUS_VERSION` in lockstep. A concurrent PR owns `check-spf.ts` and the package version, so that bump is left for the operator to sequence.


MadaBurns:
CI update — all checks green at `60af806e5`, including the four required ones (`build-and-test`, `Secret & PII scan`, `Dependency audit`, `File hygiene check`), plus `contract`, `fast-checks`, `typecheck-tests`, `dogfood-scan`, CodeQL and Semgrep. Re-runnable: `gh pr checks 990`.

This also closes the one caveat in the description: `build-and-test` runs the full root suite and **passed** (12m04s), so the local `test/wasm-integration.test.ts` failure was the missing `crates/bv-wasm-core/pkg` build artifact in a fresh worktree, not anything in this diff. That is now two independent signals rather than one diagnosis.

Issue #929, opened 2026-09-08
check_lookalikes / check_shadow_domains / discover_brand_domains: owned_by_seed at confidence 1.00 on a SHARED-PLATFORM nameserver pair (one.com) — mirror of #263/#864


MadaBurns:
## Summary

`check_lookalikes`, `check_shadow_domains` and `discover_brand_domains` all emit `ownershipVerdict: "owned_by_seed"` / `ownershipStrength: "strong"` / `combinedConfidence: 1` when the candidate and the seed share a **shared hosting platform's** nameserver pair. The only signal is `ns_set_match`. On a platform where every customer gets the same two nameservers, that is evidence of nothing: the same test "attributes" every other tenant of the platform to the seed.

This is the mirror image of #263 / #864 (same-entity domains MISSED because NS differ). Here unrelated-until-proven domains are AFFIRMED on NS alone.

Measured 2026-09-08, engine 1.36.0 / model 1.24.0 / config `a89c66e7`.

## Reproduction

Seed `net-agents.dk`, NS `ns01.one.com` / `ns02.one.com` (one.com shared hosting, Denmark).

`check_lookalikes net-agents.dk`:

```json
{"title":"Lookalike domain likely owned by same entity: net-agent.dk",
 "detail":"... net-agent.dk shares 2/2 dedicated nameservers with net-agents.dk.",
 "metadata":{"ownershipVerdict":"owned_by_seed","ownershipStrength":"strong",
             "ownershipSignals":["ns_set_match"],"attributionConfidence":"corroborated"}}
```

Note the wording "2/2 **dedicated** nameservers" — `ns01.one.com` is not dedicated to anyone.

`check_shadow_domains net-agents.dk` marks `net-agents.com`, `net-agents.eu`, `net-agents.org` as `owned_by_seed` on the same basis, and adds a "Shared NS across shadow domains ... suggesting common ownership or registrar" finding (that last clause is the honest reading, but the per-domain verdict is not hedged).

`discover_brand_domains futuresoft.dk` (classic, deep, with those names supplied as `candidate_domains`) surfaces all six at `combinedConfidence: 1` on `signals: ["ns"]` with `matchType: "set_overlap"`.

## Ground truth

[...]

Issue #939, opened 2026-09-08
ownership-attribution: unlisted uniform-set NS platforms (Hostinger, Wix, Cloud DNS, registrar defaults) still yield owned_by_seed — the class behind #929 remains open


MadaBurns:
## Summary

PR #937 (closes #929) fixed the one.com instance of a class: a nameserver set that a shared platform hands to every tenant was accepted as `owned_by_seed` evidence. The mechanism now defaults safe — any apex in `SHARED_NS_APEXES` (`src/tenants/discovery/shared-ns-hosts.ts`) can never attribute, and only `POOLED_SHARED_NS_APEXES` (Akamai) earns the complete-match arm — but the *list* still contains only what has been measured. Every uniform-set platform NOT on the list still reaches `classifyOwnership()` step 3 as "dedicated" and yields `owned_by_seed` / `strong` / confidence 1.00 on a complete match, exactly as one.com did.

From the adversarial review of #937 (step-3 analysis, quoted):

> The remaining `owned_by_seed` arms are seed-only evidence (in-bailiwick NS, the RFC 7489 §7.1 DMARC grant), the Akamai pooled match, or the **documented unlisted-platform gap**: a platform that assigns every tenant the same NS set but is not in `SHARED_NS_APEXES` still satisfies `ns_set_match` as "dedicated". Candidates: Hostinger (`ns1`/`ns2.dns-parking.com` — the default for every Hostinger domain), Wix (`ns6`/`ns7`… `.wixdns.net`), Google Cloud DNS (one of five fixed `ns-cloud-{a..e}{1..4}.googledomains.com` sets — manufacturable by any GCP account, and the existing audit pins it as "unique-per-zone", which it is not), Namecheap PremiumDNS / other registrar defaults beyond `registrar-servers.com`.

## What to do

For each candidate, measure two UNRELATED tenants over DoH before listing:

```
curl -s "https://dns.google/resolve?name=<domain-a>&type=NS"
curl -s "https://dns.google/resolve?name=<domain-b>&type=NS"
```

[...]

PR #1036, opened 2026-09-17
fix(lookalikes): publish ownership signals, and attribute pooled-nameserver ccTLD estates

MadaBurns:
> **Note on naming.** This description previously identified the real
> organisations behind the reproduction domains. Their names and the
> per-organisation security characterisations have been replaced with role
> labels (`<bank-a>`, `<bank-b>`, `<saas-a>`, `<bank-c>`), because a public PR
> is not the right place to publish a current security assessment of a named
> third party. The structural argument is unchanged; the concrete domains are
> in the branch's specs for anyone reproducing it.

## Summary

`check_lookalikes` reported `<bank-a>.co.nz` — a major bank's **genuine**
ccTLD domain — as `third_party` with `ownershipSignals: []`, i.e. "no ownership
signal links it to the seed". Reproduced against live prod.

Two separate defects were behind that one line.

## 1. The signals were computed, then thrown away

`classifyOwnership()` **did** return `['ns_shared_platform']` for the candidate.
But `buildOwnershipGateMetadata` (`ownership-attribution.ts:1031-1043`,
pre-change) emitted only `verdict` / `rationale` / `confidence` /
`severityCappedBy` — `ownershipStrength` and `ownershipSignals` existed
**solely** on `buildOwnedBySeedFinding` (`lookalike-findings.ts:137-138`).

So the empty array meant *"evaluated and not published"*, not *"never
evaluated"*. Both fields now travel on every gated finding.

## 2. The pooled-nameserver mirror of #929

#929 fixed the **uniform**-platform case (an identical NS set observes nothing
about ownership). The **pooled** case was missing.

Akamai assigns six hostnames **per zone** from a pool of ~128, so two zones
belonging to *one account* differ by design. The classifier read "different
hostnames" as "distinct infrastructure". The live rationale was: *"shares 1/6
nameservers with `<bank-a>.com` … its remaining nameservers are distinct"*
(`ownership-attribution.ts:646-651`).

New arm, deliberately narrow — seed's whole set on one pooled provider, **and**
every candidate host either on that provider or inside the candidate's own
registrable domain, **and** at least one on the provider, **and** the candidate
is an exact-label TLD variant → `unattributed` / `ns_shared_platform`.

It can **never** return `owned_by_seed`, so the info ceiling and the separate
threat observation are untouched.

## Both signals I proposed were rejected — correctly

Worth recording, because they are the obvious ideas:

- **Partial NS overlap** — rejected on the repo's own measurement.
  `shared-ns-hosts.ts:281` records an *unrelated* competitor bank sharing a
  single Akamai host with `<bank-a>.co.nz`; two 6-of-128 draws collide ~25% of
  the time. Overlap alone is not evidence.
- **Vanity NS (`ns1.<bank-a>.co.nz`) as ownership evidence** — rejected as
  **candidate-published and free to forge**, the exact class Ruling A deleted;
  and for an exact-label variant the brand-token match is tautological.
  Reframed instead as *"names no operator, so it is neither ownership nor
  distinctness evidence"* — which is what unblocks the candidate without
  introducing a forgeable signal.

The exact-label scoping (extracted `isExactLabelTldVariant`, shared with step
5d) is what preserves #929's pinned two-bank outcome. The first draft failed
that spec; the arm was narrowed rather than the pin overturned.

## Quantified, not fixed from one example

63 registered seed/candidate pairs across 20 brands' ccTLD estates
(`dig @1.1.1.1`, 2026-09-17). **Two rows move, one verdict class changes,
`owned_by_seed` count unchanged at 26.**

## Verification

| check | result |
|---|---|
| `npm run typecheck` / `lint` / `audit:oss-safety` | rc=0 |
| `npx vitest run test/check-lookalikes-pooled-platform-ns.spec.ts` | rc=0, **9/9** |
| `check-lookalikes.spec.ts` + `ownership-attribution.spec.ts` | rc=0, **108/108** |
| full `npm test` (pinned verify-capture at candidate) | **748 files / 9858 tests, exit 0** |

## ⚠️ Deliberately NOT fixed — needs your call

1. **The `<bank-a>.co.nz` observation is unchanged by this PR.** No sound
   signal makes it `owned_by_seed` on the available evidence: that ccTLD has no
   RDAP, its DMARC `rua` points at a shared vendor mailbox, and MX matches the
   seed while A does not. This PR removes the false *"no ownership signal"*
   claim; it does not change the observation either way.
2. **`<saas-a>.*` is unchanged** — its ccTLDs are on Akamai, its `.com` on
   Route 53. No shared platform, no seed-published record. Covering it needs
   both of step 5d's guards relaxed — the loosening direction that risks
   attributing an attacker's domain to the brand.
3. **The `<bank-a>.co.uk` premise is unverified** — different provider,
   different mail vendor, no `rua`. I could not establish that it belongs to
   the same organisation.
4. **SIDE DEFECT, not fixed, worth its own ticket:** one seed pair
   (`<bank-c>`) has a **byte-identical 9/9 NS set** across its `.co.uk` and
   `.com`, including 3 dedicated in-house hosts, and is *still* reported "no
   ownership signal links it". Step 3 ratios dedicated matches over the seed's
   **total** count (3/9 < 0.5) and step 4 excludes mixed sets, so a complete
   exact match falls between both arms. **3 of 63 pairs.** Left alone because
   fixing it moves a verdict **into** `owned_by_seed`, which lifts a severity
   cap — that is your decision, not an executor's.

## Housekeeping

`src/lib/ownership-attribution.ts` is now **1227 LOC** (was 1129) and trips the
oversize hook. No `bv-oversize-ok` marker was added — flagging rather than
silencing it.


MadaBurns:
> **Edited to remove third-party names.** This comment previously identified the
> real organisations behind the measured domains and quoted their live mail
> policy and registrar. Those are replaced with stable role labels
> (`<bank-a>`, `<bank-b>`, `<bank-c>`, `<industrial-a>`, `<telco-a>`,
> `<saas-a>`, `<saas-b>`) used consistently across #1036, #1039 and #1040. No
> measurement, count or conclusion changed. The concrete domains are in the
> branch's specs and the local evidence bundle.

## Three of the four open items are now measured

Posting this so the remaining decision is made against evidence rather than judgement. All three investigations were read-only; nothing here has been pushed to this branch.

### Item 3 — `<bank-a>.co.uk` premise: **HOLDS**

Verified on the wire 2026-09-18, with positive controls run **first** (`_dmarc.google.com` returned a real `rua=`, `google.com` MX resolved) so that "no rua" is a measured absence rather than a silent resolver empty:

- NS: two `*.ns.cloudflare.com` hosts
- `_dmarc.<bank-a>.co.uk` → a `p=reject` policy with strict alignment and **no `rua` tag**
- MX: a single third-party mail host · apex SPF `v=spf1 -all`

The "could not establish it is the same organisation's" half also survives a check this PR did not try. `.co.uk` **does** have RDAP (unlike `.co.nz`), but the registrar of record is a third-party domain-management firm rather than the organisation itself, the registrant is privacy-redacted, and HTTP is bot-walled (`403`, `cf-mitigated: challenge`).

Item 3 moves from *unverified* to **verified: ownership is genuinely unresolvable via DNS/RDAP/HTTP**. The decision itself is unchanged.

### Item 2 — `<saas-a>.*` / step-5d guards: decision **upheld**, but two claims in the PR body need correcting

"Loosening both guards is the unsafe direction" was asserted without measurement. It is now measured across 2,726 domains swept live, and it holds — but not evenly, and not for the stated reason.

| variant | fired | verdict-changing | confirmed FP | rate |
|---|---|---|---|---|
| base (corpus B) | 37 | 0 | 0 | — |
| guard 2 removed | 40 | 3 | 0 | 0.0% |
| guard 1 removed | 427 | 321 | 53 | **16.5%** |
| both removed | 453 | 347 | 60 | **17.3%** |
| both, against a 63-pair brand corpus | 6 | 3 | **0** | **0.0%** |

**⚠️ The last row is a false zero, and it is the important finding.** Measured against a 63-pair / 20-seed brand corpus, the loosened variant scores 0 false positives and 100% precision — because all three changes it makes there *are* the `<saas-a>.*` gap itself. The failure mode is simply absent from a 20-brand set skewed toward corporate brand-protection registrars. It took a wider stratified draw (61 labels × 44 TLDs) to expose it. Step 5d should not be re-measured against a brand-only corpus alone.

**The risk is not split between the guards.** Guard 1 (`ownership-attribution.ts:898-899`, candidate A-set must be byte-identical to the seed's) carries essentially all of it. Remove it and step 5d has no seed reference at all, degenerating into "3+ exact-label variants share one estate" — which describes a domainer portfolio as readily as a brand portfolio. Guard 2 (`:921`) produced 0 confirmed FPs, but across only 3 verdict changes: weak evidence, not a clearance.

**The stated harm is wrong.** The PR body says loosening "risks attributing an attacker's domain to the brand". Across all 674 verdict changes in every variant, **every single one resolved to `unattributed/seed_label_cohort`; not one produced `owned_by_seed`** (verified directly against the run's raw records, not just its summary). The real harm is false *exculpation* — a parked `<saas-b>.xyz`-shaped domain described as the organisation's own brand-variant portfolio. That is still a defect, and it is the mirror of the ccTLD defect this PR fixes, but it is not a severity lift.

**Carry-forward fragility in the shipped guard:** the `<saas-a>.*` estate's A records are Akamai edge-rotating — two distinct address pairs observed six minutes apart. Guard 1 and the sibling A-identity test both compare a rotating edge set, so any cross-run or slow scan will produce spurious mismatches on Akamai-fronted estates. Each single sweep was internally consistent, so same-run comparison is sound.

### Item 4 — exact-NS gap (#1039): measured, and the risk framing in that issue is **understated**

3,588 pairs constructed, 1,819 measured live. An exact-set-match arm between step 3 (`:569`) and step 4 (`:571`) produces:

- **30 verdict changes (1.65%)**, all `third_party` → `owned_by_seed`
- **0 true impersonations suppressed** — 0 of 615 impersonation-shaped pairs, 0 of 720 unrelated-label pairs
- all 30 flips come from **three seeds**: `<industrial-a>` (18), `<bank-b>` (11), `<telco-a>` (1)

The six flips the generator labelled "impersonation" are `<industrial-a>` and `<bank-b>` gTLD variants (`.net`, `.org`, `.co`) — defensive registrations sitting on the brand's own dedicated nameservers, i.e. mislabelled by the generator's stratum, not true impersonations. The Akamai component is per-**account** (a sibling brand of `<industrial-a>` draws a completely different 6-host set), so an exact match is an account-level match an attacker cannot reproduce.

**The zero is positive-controlled, not assumed.** A synthetic attacker domain — a lure-shaped label under `.example`, handed the real 9-host `<bank-b>` NS set — flips `third_party` → `owned_by_seed` under the fix, so the harness can detect the harm it reports zero of.

**⚠️ #1039 understates the blast radius.** `owned_by_seed` does not merely lift a severity ceiling — it **removes the finding entirely** (`src/tools/check-lookalikes.ts:621-626`, plus `lookalike-findings.ts:42` and `lookalike-summary-findings.ts:87`). The exposure is suppression, not severity drift. I have corrected this on #1039.

One nuance worth stating plainly: the 30-flip corpus sweep was run with the dedicated-host floor at `>= 1`. The recommendation is `>= 2`, which is validated by synthetic controls (three attacker-reachable windows — a registrar's `ns33/ns34` pair + 1 unlisted host, a second registrar's `dns1/dns2` pair + 1, and 6× akam + 1 — all close at `>= 2` and open at `>= 1`). That the two are identical on all 1,819 real pairs is the investigation's finding; I have not independently re-derived it from the logs.

### Not fixed here, but live on `main` today

Independent of this PR and of #1039, `main` **already** returns `owned_by_seed` via step 3 for unrelated domains sharing uniform self-service platforms absent from `SHARED_NS_APEXES` (83 entries): `ns1/ns2.afternic.com`, `ns5/ns6.kasserver.com` (two unrelated labels on the same pair, which is what proves uniformity), `ns1/ns2.atom.com`, `cdns1-3.interserver.net`, `ns1/ns2.digitaldays.net`. Two investigations converged on this independently from opposite directions. It is a live #939-class allowlist gap that this PR neither creates nor widens — filed separately.

Issue #1073, opened 2026-09-21
ALERT_WEBHOOK_URL is documented as a secret but deployed as a plaintext var, so every deploy prints it


MadaBurns:
## Summary

`ALERT_WEBHOOK_URL` is configured as a plaintext `vars` entry in the production config. `deploy:prod` runs `check:bindings:prod`, which runs `wrangler types --config wrangler.production.jsonc`, and that renders every plaintext var inline in its generated output. The webhook URL therefore prints **in full — including its path token — on every production deploy**, to the terminal and to any transcript or CI log capturing that output.

The URL is a capability secret: possession of it is sufficient to post to the internal ops alert endpoint. It was disclosed this way during the 3.85.0 deploy and the token is being rotated.

Also rendered by the same step, for completeness: `CF_ACCOUNT_ID`, `OWNER_ALLOW_IPS`, and the three KV namespace IDs. Those are lower-sensitivity but are disclosed by the same mechanism.

## The code already calls it a secret

`src/lib/operator-webhook-binding.ts:5-6` describes the fallback as:

> falling back through a KV last-known-good cache to the static `ALERT_WEBHOOK_URL` **secret**

So the documented intent and the deployed reality disagree. The docstring says secret; the config says var. That mismatch is the whole defect.

## Why the obvious fix does not work

`wrangler secret put ALERT_WEBHOOK_URL` fails:

```
A request to the Cloudflare API (.../secrets) failed.
Binding name 'ALERT_WEBHOOK_URL' already in use. Please use a different name and try again. [code: 10053]
```

A binding name cannot be both a var and a secret, so the var has to be removed first — and that is gated:

`scripts/inject-private-config.cjs:169`

```js
const REQUIRED_NONEMPTY_PRODUCTION_VARS = ['ALERT_WEBHOOK_URL'];
```

`validateProductionSecurityConfig` fails the injector when that var is absent or empty, so simply deleting it from `vars` blocks every deploy, fail-closed.

That gate should be **replaced, not removed**. It exists to prevent shipping with alerting silently off, which is a real failure mode worth keeping. A secret-aware equivalent (for example, asserting the name appears in `wrangler secret list`) would preserve the guarantee while allowing the value to stop being rendered.

## Priority is moderate, not urgent

[...]

PR #1076, opened 2026-09-21
fix(deploy): make the ALERT_WEBHOOK_URL prod gate secret-aware (#1073)

MadaBurns:
## Summary

Fixes the code half of #1073 — `ALERT_WEBHOOK_URL` is documented as a secret but deployed as a plaintext `vars` entry, so `wrangler types --config wrangler.production.jsonc` (run by `check:bindings:prod` inside `deploy:prod`) printed the full capability URL on every production deploy.

> ⚠️ **Draft on purpose — do not merge before reading "Merge ordering" below.** An independent review of this exact commit is in progress (board ticket SQ-122).

**What changes**

- `scripts/inject-private-config.cjs` — the var-only gate (`REQUIRED_NONEMPTY_PRODUCTION_VARS`) is **replaced, not removed**, by a secret-aware one. `assessAlertWebhookSecretGate` (pure, injectable) passes only when `wrangler secret list` shows `ALERT_WEBHOOK_URL` provisioned **and** the name is absent from `vars`. It fails closed when the name is still/also a var (the `[code: 10053]` collision state and the re-disclosure state), when the secret is missing, or when the lister errors, times out (30s) or returns unparseable output.
- One-shot `BV_ALLOW_MISSING_ALERT_SECRET=1` override for the transition deploy, following the `BV_ALLOW_STALE_SIDECARS=1` precedent. It does **not** rescue the var-still-present state.
- `wrangler.jsonc` / `wrangler.private.example.jsonc` — `ALERT_WEBHOOK_URL` removed from `vars`.
- `docs/operator-runbook.md` — new migration section with two paths: a one-deploy path using `wrangler deploy --secrets-file`, and a two-step override path. `docs/tenant-ops-runbook.md`, the `bv-mcp-operations` skill and `CHANGELOG.md` updated to match.
- Runtime is unchanged: a secret and a var are both plain `env` properties, and `resolveAlertWebhookUrl` precedence (dynamic `BV_WEB` value wins) is untouched.

Wrangler flags were read from the **installed** binary (4.131.1), not from memory: `secret list --name --format json`, `deploy`/`versions upload --secrets-file`, and `secret put` reading stdin when not a TTY.

Deliberately not done: `ALERT_WEBHOOK_URL` was not added to `PRODUCTION_REQUIRED_SECRETS` — that built-in gate has no override, so listing the secret there before it is provisioned would turn the transition into a deploy outage.

## Merge ordering

Once this merges, the next `deploy:prod` **fails closed** until the operator has (a) removed `ALERT_WEBHOOK_URL` from the private overlay's `vars` and (b) provisioned it as a Worker secret — see the new runbook section. That is the intended guarantee ("prod never ships with alerting silently off"), but it means the Cloudflare-side move is an operator step that has to be planned with this merge. No Cloudflare state was touched by this PR or by the agent that wrote it. The leaked token rotation mentioned in #1073 is likewise operator-side.

## Type of change

- [x] Bug fix
- [x] Security fix
- [x] CI / tooling

## Security impact

Yes — deploy-gate and secret handling. Removes a capability URL from rendered deploy output; keeps the alerting-configured guarantee fail-closed. Read-only audit of the remaining plaintext var **names** found nothing else that reads as a credential or capability URL (`CF_ACCOUNT_ID`, `OWNER_ALLOW_IPS` and the KV/D1 IDs remain, as already noted in #1073 as lower-sensitivity).

## Test plan

- [x] Unit tests added / updated — 14 behavioural tests on the gate and the lister via an injected fake `spawnSync`/lister (no subprocess, no network)
- [x] `npx vitest run test/audits/private-config-injection.node.test.ts test/operator-webhook-binding.spec.ts` → 40 passed
- [x] `npm run typecheck` passes — and the generated `Env` no longer contains `ALERT_WEBHOOK_URL` as a var
- [x] `npm run audit:oss-safety` → 69 passed
- [x] Negative control: neutering the var-present check produces 3 genuine assertion failures on exactly that behaviour
- [ ] `npm test` — full suite left to CI
- [ ] Manual testing against a real Cloudflare account — **not performed** (agent was barred from any Cloudflare mutation)

## Checklist

- [x] Changes follow existing code conventions
- [x] No secrets, credentials, or internal references included
- [x] Error messages follow the safe-prefix convention (n/a — deploy-script output only)
- [x] New tool? n/a


MadaBurns:
CI note: the red `build-and-test` here is the **CI Docs** job, and it is a real failure, not a flake. `test/audits/security-capability-inventory.node.test.ts:39` reports `expected [ 'ALERT_WEBHOOK_URL' ] to deeply equal []` — this PR adds `ALERT_WEBHOOK_URL` to the operator secret-reconstruction block in `docs/operator-runbook.md`, and that audit requires every name there to be classified in `MCP_SECURITY_CRITICAL_SECRET_KEYS` (`src/lib/security-capabilities.ts`). It is not yet.

Planned repair: classify it in that inventory (it *is* a capability — the premise of #1073), after checking the list's two runtime consumers. It will land as a follow-up commit on this branch together with anything the in-progress independent review turns up, so the branch is repaired once. Staying in draft until then.

MadaBurns:
Independent review of this exact commit (`1d59da5f1`) is back.

**Gate logic: no fail-open path found.** Eight of nine probed classes are safe — malformed or hostile `wrangler secret list` output fails closed, secret-name matching is exact, the worker name always comes from `wrangler.jsonc` and an undefined name fails closed, the var-present check always runs before the override check, the gate reads the fully merged `vars`, all three injector callers propagate a non-zero exit, the tests assert on real gate output, and runtime `env` access is unchanged.

**One confirmed defect, docs-only:** the runbook's recommended one-deploy `--secrets-file` path says no override is needed. That is wrong — `deploy:prod` runs the injector (and its `wrangler secret list` check) *before* the final `wrangler deploy` step that creates the secret, so that path fails closed every time. It needs `BV_ALLOW_MISSING_ALERT_SECRET=1` on that one command, with the following deploy run without it as the proof the secret landed.

The candidate was rejected on the board for that plus the inventory-audit CI failure above. A replacement candidate that reuses the reviewed gate code unchanged and fixes both is in progress; it will arrive as a new PR that supersedes this one, and this draft will be closed with a pointer.

MadaBurns:
Superseded by #1077, which carries this gate script, its tests and both wrangler configs byte-identical, and fixes the two defects recorded above (the runbook's one-deploy path, and the unclassified `ALERT_WEBHOOK_URL` that failed the security-capability-inventory audit). Closing this draft unmerged; the branch is left in place for reference.

